PR #8, opened 2023-03-31
urler: use --set-* and --show-* for the different components

bagder:
For consistency


bagder:
Or should it be `--get-*` ?

PR #14, opened 2023-03-31
Makefile: create man1 directory if missing; install manpage to DESTDIR

ZhongRuoyu:
Currently, installation will fail if `$(MANDIR)/man1/` does not exist. Fix that by creating the directory first.

Also, since the manpage is a `install` target, make sure it follows `DESTDIR`.


bagder:
Thanks!

PR #17, opened 2023-03-31
Makefile: add prefix variable to support tools like homebrew

marianposaceanu:
Update the Makefile to add support for a prefix in order support tools like Homebrew.


bagder:
Thanks!

dfandrich:
$DESTDIR and $prefix have different functions, and both might be used.

bagder:
True. We should fix this to support both properly.

PR #19, opened 2023-03-31
add error handling for when no arguments passed for options at the end of command

jacobmealey:
master currently segfaults if you run `urler --url`, this is a quick fix so let me know if there should be more careful considerations taken. Note that this is different from #16 not sure what is different from testing but mine still seg faults if no URL passed.

### DIFF
--- a/urler.c
+++ b/urler.c
@@ -279,11 +279,8 @@ int main(int argc, const char **argv)
     else if(argv[0][0] == '-' && argv[0][1] == '-') {
       /* dash-dash prefixed */
       int usedarg = 0;
-      if(argc < 2) {
-          help("Invalid Arguments passed");
-          return -1;
-      }
-      if(getlongarg(&o, argv[0], argv[1], &usedarg))
+      
+      if(argc < 2 || getlongarg(&o, argv[0], argv[1], &usedarg))
         help("unknown option");
 
       if(usedarg) {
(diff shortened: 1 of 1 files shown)


urler.c
```diff
@@ -279,7 +279,8 @@ int main(int argc, const char **argv)
     else if(argv[0][0] == '-' && argv[0][1] == '-') {
       /* dash-dash prefixed */
       int usedarg = 0;
-      if(getlongarg(&o, argv[0], argv[1], &usedarg))
+      
+      if(argc < 2 || getlongarg(&o, argv[0], argv[1], &usedarg))
```
bagder, line 283: Doesn't this prevent the options that don't take arguments from working if specified last?

jacobmealey: Oh woops! my bad. I just pushed another revision that should work better. I have it checking if the argument is null in `getlongarg()` It looks like this is something that should be checked for all arguments that take options. for example `urler --url github.com --append-path`

jacobmealey:
Okay, I added checks for when the user runs commands that end with `--append-path` `--append-url` and `--url` but don't provide text. 

ps sorry for the many commits and if this PR is noisy

bagder:
This still leaves out several other `---set-*` options from the check. I propose we maybe do it this way instead: #21

bagder:
Lots of other internal changes but this was fixed in f64e450612740a1

PR #24, opened 2023-04-01
urler: it does not need a full URL, but enough components set to make…

bagder:
… one

Setting host + scheme is enough.


bagder:
This partly reverts e2e1b7d7eb2a8dd04f but brings back previous functionality.

PR #26, opened 2023-04-02
README: add reference to required dependencies

ryru:
It may be obvious for a curl project but I think the hint is still not wrong.


bagder (approved):

bagder:
Thanks!

PR #29, opened 2023-04-02
feat(test.py)

sevehub:
test.pl equivalent python script extended for Windows/Mingw64-w64 toolchain.


bagder:
But why do we need the tests run by python?

sevehub:
I think it's just more convenient to have a Python test script when you compile the program on Windows (Mingw64-w64).
Decided to share it in case you need it.

bagder:
Doesn't it run perl scripts just as easily?

I think we should separate the test data from the test runner, then I'm fine with having it in python if that is easier for some reason. But I don't think we need two separate ones.

sevehub:
I am sure it does, but I have no idea how to write platform-independent Perl scripts.

I can move the test urls inside a csv file (test.csv), haven't seen commas used within the program's options.
Better close this PR and submit a new one.

bagder:
I think we should expand the test file format at once. It also needs to be able to verify the exit code and multi-line outputs.

sevehub:
Multilines are still single lines with a bunch of line separators in them.
Format can be something like:
input, output, exit code
--url https://curl.se?name=hello --append query=search=string, https://curl.se/?name=hello&search=string, 0

sevehub:
Second thought, I have seen the ongoing PR.
We can copy the test data inside a JSON file, would it be possible to have a dedicated branch?

bagder:
> We can copy the test data inside a JSON file

As long as the tests remain easy to make and read, I don't mind JSON.

> would it be possible to have a dedicated branch?

Please elaborate. Why do we want a branch?

sevehub:
It's ok, I will use the master branch.
I close this commit meanwhile.

Issue #49, opened 2023-04-03
multiple URL JSON output?


bagder:
How is the JSON format supposed to look like when trurl outputs multiple objects?

Right now it looks like this when given 4 URLs:
~~~
$ cat urls.txt | ./trurl --json -f -
{
  "url": "http://localhost/",
  "scheme": "http",
  "host": "localhost",
  "port": "80",
  "path": "/",
}
{
  "url": "http://haxx.se/",
  "scheme": "http",
  "host": "haxx.se",
  "port": "80",
  "path": "/",
}
{
  "url": "https://curl.se/",
  "scheme": "https",
  "host": "curl.se",
  "port": "443",
  "path": "/",
}
{
  "url": "http://curl.haxx.se/",
  "scheme": "http",
  "host": "curl.haxx.se",
  "port": "80",
  "path": "/",
}

paulroub:
Good question - as the code is written, it would take some gymnastics to make it behave differently for one URL (where a single `{ ... }` would do) vs multiple (where you'd want an array for things to be parseable by anything else). 

There's probably an advantage to always outputting an array, even if it's a single URL -- consistent parsing, consistent examples, etc.  

So I'd suggest:

```json
[
  {
    "url": "http://localhost/",
    "scheme": "http",
    "host": "localhost",
    "port": "80",
    "path": "/"
  }
  {
    "url": "http://haxx.se/",
    "scheme": "http",
    "host": "haxx.se",
    "port": "80",
    "path": "/"
  }
  {
    "url": "https://curl.se/",
    "scheme": "https",
    "host": "curl.se",
    "port": "443",
    "path": "/"
  }
  {
    "url": "http://curl.haxx.se/",
    "scheme": "http",
    "host": "curl.haxx.se",
    "port": "80",
    "path": "/"
  }
]
```

(dropping the trailing commas, which are not valid JSON)

bagder:
Thanks, I will poke the code accordingly!

PR #52, opened 2023-04-03
trurl: output JSON as an array

bagder:
To properly work with many URLs

Fixes #49

PR #53, opened 2023-04-03
One approach to parseable JSON output

paulroub:
Addresses multiple-URL JSON output (and removal of trailing commas in JSON) mentioned in #49


paulroub:
Reopening this, as the sample tests might be of use.

bagder:
#52 was merged with the test code from here.

PR #56, opened 2023-04-03
Correct JSON example in README

paulroub:
Updated JSON example to reflect the new JSON behavior (always an array, no trailing commas).


bagder (approved):
There's a dupe in the man page that should be updated as well...

PR #60, opened 2023-04-04
trurl: add --trim

### LINKED ISSUE #50: suggestion: strip parameters
something i do a lot, remove tracking url parameters before sharing a url. i see `--append` but it'd be cool to have something to strip the garbage.

bagder:
--trim can remove (parts of) components. Starting out with query support for removing specific "tuples".

Man page and README updates. Added test cases.

Fixes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,9 @@ command line tool for URL parsing and manipulation
       "fragment": "frag"
     }
   ]
+
+  $ trurl "https://example.com?search=hello&utm_source=tracker" --trim query="utm_*"
+  https://example.com/?search=hello
 ~~~
 
 ## Install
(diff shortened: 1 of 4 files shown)


ijoseph:
This is so powerful! We can now do something like: 

```bash
    trimmed=$(trurl "$url" --trim query='utm_*')
    trimmed=$(trurl "$trimmed" --trim query='fbclid*')
    trimmed=$(trurl "$trimmed" --trim query='notif*')
    trimmed=$(trurl "$trimmed" --trim query='ref*') 
```

It's a bit whack-a-mole, but like with AdBlock, lists could be crowdsourced. Love it.

bagder:
You can also do any amount of `--trim` actions on the same command line.

PR #65, opened 2023-04-04
Use Test::More inside of test.pl

oalders:
This shouldn't require any dependency management, since Test::More has
been provided with Perl since Perl release 5.6.2

perl -MModule::CoreList -E 'print Module::CoreList->first_release('Test::More');'
5.006002


bagder (approved):

bagder:
Thanks!

PR #68, opened 2023-04-04
Add an INSTALL variable to the Makefile

xfgusta:
This is useful to distribution packagers who may wish to set

```sh
INSTALL='install -p'
```

in order to preserve timestamps from the source tarball.

https://docs.fedoraproject.org/en-US/packaging-guidelines/#_timestamps


bagder (approved):

bagder:
Thanks!

PR #81, opened 2023-04-05
Update README.md

DrainGangScholar:
you were missing the letter g from the gcc calls


bagder:
Not really. That's a screendump from an actual build. gcc is usually also working as "cc" for old legacy reasons.

DrainGangScholar:
Sorry, I didn't know that!

Issue #76, opened 2023-04-05
When given a file of URLs, continue even if one is bad


bagder:
This case should output four URLs - there is one bad one in the middle.

~~~
$ cat urls.txt 
localhost
haxx.se
ftp://ex ample.com
https://curl.se
curl.haxx.se
$ trurl -f urls.txt
http://localhost/
http://haxx.se/
trurl note: Bad hostname [ftp://ex ample.com]
trurl error: not enough input for a URL
trurl error: Try trurl -h for help

bharsaklemukesh975:
can I work on this? I would like to start my first open source contribution with this .
Thanks

jay:
@havbon has already submitted a fix in #84

jay:
If a user passes in 5 lines I would think they expect to get 5 lines back, so what about if bad URLs just result in blank lines

PR #84, opened 2023-04-05
ignore bad urls when given a file of URLs

havbon:
for #76 

I have developed this on windows 11.

the output is now:
`
PS C:\Users\Håvard\git\trurl> .\trurl.exe --url-file .\trurl\urls.txt
http://localhost/
http://haxx.se/
https://curl.se/
http://curl.haxx.se/
`

urls.txt:
`
PS C:\Users\Håvard\git\trurl> type .\trurl\urls.txt
localhost
haxx.se
ftp://ex ample.com
https://curl.se
curl.haxx.se
`

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -598,9 +598,14 @@ static void singleurl(struct option *o,
                      (o->accept_space ?
                       (CURLU_ALLOW_SPACE|CURLU_URLENCODE) : 0));
       if(rc) {
-        if(o->verify)
-          errorf(ERROR_BADURL, "%s [%s]", curl_url_strerror(rc), url);
-        warnf("%s [%s]", curl_url_strerror(rc), url);
+        if (o->url) { // check if we are reading from a file
+          return;
+        } 
+        else {
+          if(o->verify)
+            errorf(ERROR_BADURL, "%s [%s]", curl_url_strerror(rc), url);
+          warnf("%s [%s]", curl_url_strerror(rc), url);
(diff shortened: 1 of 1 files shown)


trurl.c
```diff
@@ -598,9 +598,14 @@ static void singleurl(struct option *o,
...
-          errorf(ERROR_BADURL, "%s [%s]", curl_url_strerror(rc), url);
-        warnf("%s [%s]", curl_url_strerror(rc), url);
+        if (o->url) { // check if we are reading from a file
+          return;
+        } 
+        else {
+          if(o->verify)
+            errorf(ERROR_BADURL, "%s [%s]", curl_url_strerror(rc), url);
+          warnf("%s [%s]", curl_url_strerror(rc), url);
+        }
```
bagder, line 608: I was going to say something about how this changes behavior for when URLs are provided via file or when listed on the command line, but I also have second thoughts about how to behave on bad URLs on the command line.

I think we should rather abort the process (for that instance) if the URL cannot be parsed. Wherever that URL comes from.

havbon: You are probably right, what about a --ignore-bad-url flag or something like that? And then the default without that flag is throwing error on any bad url as you said.

bagder: I propose my take in #91: it makes it just ignore bad URLs and move on to the next, as we already have `--verify` that will abort with an error on the first bad URL.

Issue #67, opened 2023-04-04
Cannot compile on Ubuntu 20.04.5 LTS


baykovr:
(Hello Daniel, thank you for the project)

I cloned trurl and fetched `libcurl4-openssl-dev` on Ubuntu Focal (20.04) and found that `libcurl4-openssl-dev` (7.68.0-1ubuntu2.18) alone is insufficient.

I see the note in the readme on CURLUPART_ZONEID, is it the case that the minimum required version for compiling is thus 7.81.0?

```
~/src/github.com/trurl (master ✔) make 
cc  -W -Wall -pedantic -g   -c -o trurl.o trurl.c
trurl.c: In function ‘get’:
trurl.c:365:18: error: ‘CURLUE_NO_ZONEID’ undeclared (first use in this function); did you mean ‘CURLUPART_ZONEID’?
  365 |             case CURLUE_NO_ZONEID:
      |                  ^~~~~~~~~~~~~~~~
      |                  CURLUPART_ZONEID
trurl.c:365:18: note: each undeclared identifier is reported only once for each function it appears in
trurl.c:369:55: warning: implicit declaration of function ‘curl_url_strerror’; did you mean ‘curl_multi_strerror’? [-Wimplicit-function-declaration]
  369 |               fprintf(stderr, PROGNAME ": %s (%s)\n", curl_url_strerror(rc),
      |                                                       ^~~~~~~~~~~~~~~~~
      |                                                       curl_multi_strerror
trurl.c:47:25: warning: format ‘%s’ expects argument of type ‘char *’, but argument 3 has type ‘int’ [-Wformat=]
   47 | #define PROGNAME        "trurl"
      |                         ^~~~~~~
trurl.c:369:31: note: in expansion of macro ‘PROGNAME’
  369 |               fprintf(stderr, PROGNAME ": %s (%s)\n", curl_url_strerror(rc),
      |                               ^~~~~~~~
trurl.c:369:44: note: format string is defined here
  369 |               fprintf(stderr, PROGNAME ": %s (%s)\n", curl_url_strerror(rc),
      |                                           ~^
      |                                            |
      |                                            char *
      |                                           %d
make: *** [<builtin>: trurl.o] Error 1
```

bagder:
This is mentioned in the README. We need to add some #ifdef magic to handle older libcurl versions proper.

bagder:
I believe this is fixed now.

PR #85, opened 2023-04-05
trurl: add support for libcurl versions below 7.80.0

baykovr:
This patch allows compiling with versions of libcurl as low as 7.68.0 (the latest available to Ubuntu 20.04 LTS #67). 

This approach gets us building, with okay test coverage but it should be better. @bagder I'm not great at C but happy to learn and contribute, could you advise on how the implementation may be improved?

### Tests
The test coverage fails 2 of 49.
```
not ok 28 - ./trurl --url https://10.1/we/are.html --get "{host}"
#   Failed test './trurl --url https://10.1/we/are.html --get "{host}"'
#   at ./test.pl line 66.
#          got: '10.1'
#     expected: '10.0.0.1'
not ok 29 - ./trurl --url https://[fe80::0000:20c:29ff:fe9c:409b]:8080/we/are.html --get "{host}"
#   Failed test './trurl --url https://[fe80::0000:20c:29ff:fe9c:409b]:8080/we/are.html --get "{host}"'
#   at ./test.pl line 66.
#          got: '[fe80::0000:20c:29ff:fe9c:409b]'
#     expected: '[fe80::20c:29ff:fe9c:409b]'
```

### Warns
```
trurl.c: In function ‘get’:
trurl.c:51:25: warning: format ‘%s’ expects argument of type ‘char *’, but argument 3 has type ‘int’ [-Wformat=]
   51 | #define PROGNAME        "trurl"
      |                         ^~~~~~~
trurl.c:387:31: note: in expansion of macro ‘PROGNAME’
  387 |               fprintf(stderr, PROGNAME ": %s (%s)\n", curl_url_strerror(rc),
      |                               ^~~~~~~~
trurl.c:387:44: note: format string is defined here
  387 |               fprintf(stderr, PROGNAME ": %s (%s)\n", curl_url_strerror(rc),
      |                                           ~^
      |                                            |
      |                                            char *
      |                                           %d
trurl.c:383:13: warning: case value ‘18’ not in enumerated type ‘CURLUcode’ {aka ‘enum <anonymous>’} [-Wswitch]
  383 |             case CURLUE_NO_ZONEID:
      |             ^~~~
cc   trurl.o  -lcurl -o trurl
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,7 @@
 TARGET = trurl
 OBJS = trurl.o
 LDLIBS = -lcurl
-CFLAGS := $(CFLAGS) -W -Wall -pedantic -g \
-	-D LIBCURL_VERNUM="$(shell curl-config --vernum)"
-
+CFLAGS := $(CFLAGS) -W -Wall -pedantic -g
 MANUAL = trurl.1
 
 PREFIX ?= /usr/local
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 TARGET = trurl
 OBJS = trurl.o
 LDLIBS = -lcurl
-CFLAGS := $(CFLAGS) -W -Wall -pedantic -g
+CFLAGS := $(CFLAGS) -W -Wall -pedantic -g \
+	-D LIBCURL_VERNUM="$(shell curl-config --vernum)"
+
 MANUAL = trurl.1
 
 PREFIX ?= /usr/local
--- a/trurl.c
+++ b/trurl.c
@@ -51,7 +51,7 @@
 #define PROGNAME        "trurl"
 
 /* if libcurl < 7.80.0 */
-#if (defined(LIBCURL_VERNUM) && (LIBCURL_VERNUM < 0x075000))
+#if LIBCURL_VERSION_NUM < 0x075000
 #define CURLUE_NO_ZONEID 18
 #define CURLU_ALLOW_SPACE 0
 #define curl_url_strerror(error) \
--- a/trurl.c
+++ b/trurl.c
@@ -59,7 +59,7 @@
    here such that we may at least show the numeric value the user 
    may lookup */
 #define curl_url_strerror(error) \
-  "unsupported CURLUcode error"
+  fprintf(stderr, "URL error: %u\n", (int)error)
 #endif
 
 struct var {
@@ -340,7 +340,7 @@ static int getarg(struct option *op,
   else if(!strcmp("--accept-space", flag))
     if(!CURLU_ALLOW_SPACE) {
       warnf("%s","--accept-space is not supported in this version of libcurl");
-    }
+    } 
     else {
       op->accept_space = true;
     }
--- a/trurl.c
+++ b/trurl.c
@@ -54,8 +54,12 @@
 #if LIBCURL_VERSION_NUM < 0x075000
 #define CURLUE_NO_ZONEID 18
 #define CURLU_ALLOW_SPACE 0
+/* curl_url_strerror is not available below this version number, 
+   we need a safe way to convert CURLUcode into a const char*
+   here such that we may at least show the numeric value the user 
+   may lookup */
 #define curl_url_strerror(error) \
-  fprintf(stderr, "URL error: %s\n", curl_easy_strerror(error))
+  "unsupported CURLUcode error"
 #endif
 
 struct var {
@@ -334,7 +338,12 @@ static int getarg(struct option *op,
   else if(!strcmp("--verify", flag))
     op->verify = true;
   else if(!strcmp("--accept-space", flag))
-    op->accept_space = true;
+    if(!CURLU_ALLOW_SPACE) {
+      warnf("%s","--accept-space is not supported in this version of libcurl");
+    }
+    else {
+      op->accept_space = true;
+    }
   else if(!strcmp("--sort-query", flag))
     op->sort_query = true;
   else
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -1,7 +1,9 @@
 TARGET = trurl
 OBJS = trurl.o
 LDLIBS = -lcurl
-CFLAGS := $(CFLAGS) -W -Wall -pedantic -g
+CFLAGS := $(CFLAGS) -W -Wall -pedantic -g \
+	-D LIBCURL_VERNUM="$(shell curl-config --vernum)"
```
bagder, line 5: `LIBCURL_VERSION_NUM` is already provided in the public headers. No need for any magic shell tricks to get it!

trurl.c
```diff
@@ -50,6 +50,14 @@
 
 #define PROGNAME        "trurl"
 
+/* if libcurl < 7.80.0 */
+#if (defined(LIBCURL_VERNUM) && (LIBCURL_VERNUM < 0x075000))
+#define CURLUE_NO_ZONEID 18
+#define CURLU_ALLOW_SPACE 0
+#define curl_url_strerror(error) \
+  fprintf(stderr, "URL error: %s\n", curl_easy_strerror(error))
```
bagder, line 58: curl_easy_strerror() cannot do the same job, they work on different number ranges

trurl.c
```diff
@@ -50,6 +50,14 @@
 
 #define PROGNAME        "trurl"
 
+/* if libcurl < 7.80.0 */
+#if (defined(LIBCURL_VERNUM) && (LIBCURL_VERNUM < 0x075000))
+#define CURLUE_NO_ZONEID 18
+#define CURLU_ALLOW_SPACE 0
```
bagder, line 56: This makes it build, but  `--accept-space` will not work. Shouldn't we then output some warning or something is that option is still used?

bagder (changes requested):
Thanks a lot for this. I had some minor comments on it...

Makefile
baykovr: Fixed! Thanks for the tip.

trurl.c
baykovr: In cases --accept-space is specified, and `CURLU_ALLOW_SPACE` is false we now emit a warning and disable the accept-space behavior (sample below). This results in test cases breaking, but I can fix them up as well (perhaps to detect the warning in the pl). 

Thoughts on the implementation?
```
./trurl  --accept-space --url "gopher://localhost/ with space"  -g '{path}'
trurl note: --accept-space is not supported in this version of libcurl
/ with space
```

trurl.c
baykovr: This one was very challenging, my goal was to at last emit the numeric error code to the user such that they may look it up, however I am not clever enough to solve this within the macro!

I tried the following
```C
#define curl_url_strerror(error) \
fprintf(stderr, "CURLUcode: %d\n", error)
```
 but this results in a segfault since we are not actually returning a `const char*`.

I then tried more clever tricks but hit `ISO C forbids braced-groups within expressions` which would harm portability ... and lastly worry that I will end up introducing a buffer overflow or some other terrible thing with other approaches.

I thus settled on a static error message, which I am unhappy with since we will eat the error code in such a case - it's safe and it works but it's not great. Thoughts?
```
./trurl ':<>https!!??illegal/https://foo.com'                               
trurl note: unsupported CURLUcode error [:<>https!!??illegal/https://foo.com]
trurl error: not enough input for a URL
trurl error: Try trurl -h for help
```

baykovr:
(Will rebase and fix conflicts once we agree on an approach - happy to revise this over and over until it's well done.)

trurl.c
```diff
@@ -54,8 +54,12 @@
 #if LIBCURL_VERSION_NUM < 0x075000
 #define CURLUE_NO_ZONEID 18
 #define CURLU_ALLOW_SPACE 0
+/* curl_url_strerror is not available below this version number, 
+   we need a safe way to convert CURLUcode into a const char*
+   here such that we may at least show the numeric value the user 
+   may lookup */
 #define curl_url_strerror(error) \
-  fprintf(stderr, "URL error: %s\n", curl_easy_strerror(error))
+  "unsupported CURLUcode error"
```
bagder, line 62: I think it should rather be `fprintf(stderr, "URL error: %u\n", (int)error);`

trurl.c
```diff
@@ -322,7 +326,13 @@ static int getarg(struct option *op,
   else if(!strcmp("--verify", flag))
     op->verify = true;
   else if(!strcmp("--accept-space", flag))
-    op->accept_space = true;
+    if(!CURLU_ALLOW_SPACE) {
+      warnf("%s","--accept-space is not supported in this version of libcurl");
+      op->accept_space = false;
```
bagder, line 331: No need to assign it to false when it already is false and will remain so.

baykovr: Fixed

trurl.c
baykovr: Fixed

trurl.c
bagder: Ah right, you would have to provide a fake version of `curl_url_strerror()` that returns the proper string. Like:
~~~c
static char *fake_strerror(CURLUcode error)
{
  static buffer[128];
  curl_msnprintf(buffer, sizeof(buffer), "CURLUcode: %d\n", error);
  return buffer;
}

bagder:
@baykovr please take a new look at how master runs for you and see if most issues are not already fixed

bagder:
I believe this is already addressed. If there are remaining issues, please let us know!

baykovr:
> @baykovr please take a new look at how master runs for you and see if most issues are not already fixed

Works well, thanks.

PR #87, opened 2023-04-05
Nicer JSON tests and test output

paulroub:
Given a minimum 12-year-old Perl version, that is.

Perl's `JSON::PP` module (pure-Perl JSON handler) has been part of the standard module list since 5.13.9, released in January 2011.

If we require that version:

- the tests are nicer to read and write
- the tests decode the output, thereby *requiring* trurl to output valid JSON
- the failure output is data-aware and more readable


bagder (approved):
Nice!

PR #95, opened 2023-04-06
Makefile: add uninstall target, add check before creating directories

sm3vhf:
Add a uninistall target to remove binaries (if they have been installed) Add a check before creating the BINDIR and MANDIR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,14 @@ $(TARGET): $(OBJS)
 trurl.o:trurl.c version.h
 
 install:
-	$(INSTALL) -d $(BINDIR)
+	[ -d $(BINDIR) ] || $(INSTALL) -d $(BINDIR)
 	$(INSTALL) -m 0755 $(TARGET) $(BINDIR)
-	$(INSTALL) -d $(MANDIR)
+	[ -d $(MANDIR) ] || $(INSTALL) -d $(MANDIR)
 	$(INSTALL) -m 0644 $(MANUAL) $(MANDIR)
+
+uninstall:
+	[ -f $(BINDIR)/$(TARGET) ] && rm $(BINDIR)/$(TARGET)
+	[ -f $(MANDIR)/$(MANUAL) ] && rm $(MANDIR)/$(MANUAL)
 
 clean:
 	rm -f $(OBJS) $(TARGET)


Makefile
```diff
@@ -15,11 +15,15 @@ $(TARGET): $(OBJS)
 trurl.o:trurl.c version.h
 
 install:
-	$(INSTALL) -d $(BINDIR)
+	[ -d $(BINDIR) ] || $(INSTALL) -d $(BINDIR)
 	$(INSTALL) -m 0755 $(TARGET) $(BINDIR)
-	$(INSTALL) -d $(MANDIR)
+	[ -d $(MANDIR) ] || $(INSTALL) -d $(MANDIR)
```
emanuele6, line 20: This is pointless, `install` does nothing if the argument directory already exists as a directory.
And also slightly wrong, since now if a user tries to run `make install` without root permissions, `install` will not complain with `install: cannot change permissions of ‘bar’: Operation not permitted` and will just carry on running the next command.
I suggesting just running `install -d`.

Makefile
```diff
@@ -15,11 +15,15 @@ $(TARGET): $(OBJS)
...
 	$(INSTALL) -m 0644 $(MANUAL) $(MANDIR)
 
+uninstall:
+	[ -f $(BINDIR)/$(TARGET) ] && rm $(BINDIR)/$(TARGET)
+	[ -f $(MANDIR)/$(MANUAL) ] && rm $(MANDIR)/$(MANUAL)
```
emanuele6, line 25: I suggest just using the classic `rm -f`. `rm -f` will not complain if the file to delete doesn't exist.
Also `[ -f ]` is problematic it will cause `make uninstall` to not delete broken symlinks, `[ -e ]` has the same problem.

Makefile
sm3vhf: It didn't continue for me....

$ LANG=en_EN make install
install -d /usr/local/bin
install: cannot change permissions of '/usr/local/bin': Operation not permitted
make: *** [Makefile:18: install] Error 1

$ ls -ald /usr/local/bin
drwxrwxr-x 3 root staff 4096 11 apr 07.07 /usr/local/bin

$ groups
--cut-- staff --cut--

PR #99, opened 2023-04-06
flush stdout after handling every URL

### LINKED ISSUE #92: Maybe add an option to disable output buffering
One possible fun use case for `trurl` could be:
```sh
# server to which clients can send one line
get_line_from_port 12345 |
# trurl filters out invalid URLs, and outputs normalised URLs
trurl -f - |
# valid URLs are logged
tee /var/downloadlog |
# do something with the URLs
download_urls_to /var/public/downloads
```
or:
```sh
# server to which clients can send one line
get_line_from_port 12345 |
# trurl filters out invalid URLs, and adds something to valid URLs
trurl -f - -s 'user=me' -s 'password=hunter2' -a 'query=source=myservice' |
# do something with the modified and validated URLs
send_requests
```
or:
```sh
# server to which clients can send one line
get_line_from_port 12345 |
# trurl parse the URLs as JSON
trurl -f - --json |
# jq parses the JSON, and only prints URLs that are for port 2023
jq -nr --unbuffered --stream 'fromstream(1 | truncatestream(inputs)) | select(.port == 2023).url' |
# do something with the URLs
do_something
```

But one reason why this doesn't work very well is that `trurl` output buffers when printing to pipes by default \(since it uses `stdio` functions\):
```sh
bash-5.1$ { printf %s\\n hi hello; sleep 2; printf %s\\n hey ;} | trurl -f - | cat
<2 seconds wait>
http://hi/
http://hello/
http://hey/
bash-5.1$
```

Luckly, if you want to use `curl` with for of those usecase, at least in GNU systems, you can use `stdbuf` to change `ŧrurl`'s stdout's buffering mode to line buffering \(or no buffering with `-o0` instead of `-oL`\):
```sh
bash-5.1$ { printf %s\\n hi:20 hello:500; sleep 2; printf %s\\n hey:40 ;} | stdbuf -oL trurl -f -
http://hi/
http://hello/
<2 seconds wait>
http://hey/
bash-5.1$
```

It would be probably nicer if there was an option users could use to make this work portably and reliably, that either enables line buffering or no buffering on stdout \(with `stdvbuf()`\), or that just flushes stdout after every URL is printed or `-g` is evaluated or object of `--json` output is fully printed.

~~Another reason why this doesn't work very well is that the presense of an invalid URL \(e.g. `\`\) in the input causes `trurl` to stop, and return failure immediately, but if I recall correctly it was mentioned in the presentation video that this will be changed in the future.~~ \(solved by 21253394fa9ecc68d9b838ff03a27e5eda63f9ae\)

emanuele6:
Fixes #92

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -731,6 +731,8 @@ static void singleurl(struct option *o,
       }
     }
 
+    fflush(stdout);
+
     freeqpairs();
 
     o->urls++;


bagder:
Is there a compelling reason we don't just do this always unconditionally?

emanuele6:
Not really. We could just always flush if we don't mind the performance penalty for large inputs.

bagder:
I think we should rather take that path - for now at least.

bagder (approved):

bagder:
Thanks!

Issue #90, opened 2023-04-06
split the query in JSON


bagder:
The #44 proposal for JSON output shows the query component being split up in sub-components. That's was not implemented in the first json take but is probably still a good idea! Probably *in addition* to the full query. Something like:
~~~
$ trurl --url "https://curl.se/user?id=foo&size=44" --json
{
  "host": "curl.se",
  "scheme": "https",
  "port": "443",
  "path": "user",
  "query": "id=foo&size=44",
  "q": {
    "id": "foo",
    "size": "44"
  } 
}
~~~

Leaves a few questions: what if there are many `id` or `size` ? What if there is no `=` at all?

xfgusta:
> Leaves a few questions: what if there are many `id` or `size` ?

What about creating an array that preserves the order of the query parameters?

**`id=foo&id=bar&size=44`**

```json
{
  "q": {
    "id": ["foo", "bar"],
    "size": "44"
  }
}
``` 

> What if there is no = at all?

I think it's best to represent it with an empty string.

**`id=foo&size`**

```json
{
  "q": {
    "id": "foo",
    "size": ""
  }
}
```

bagder:
Makes perfect sense!

Since I merged 9a7cc371f6c, this should be fairly straight forward to implement.

bagder:
can we do something nicer than `"q"` there?

bagder:
#100 is a first take, not quite there yet.

xfgusta:
> can we do something nicer than `"q"` there?

I think we could use `params`:

```
$ trurl --url "https://curl.se/user?id=foo&size=44" --json
{
  "host": "curl.se",
  "scheme": "https",
  "port": "443",
  "path": "user",
  "query": "id=foo&size=44",
  "params": {
    "id": "foo",
    "size": "44"
  } 
}
```

Note that `params` is used in the JavaScript [`URLSearchParams` API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

bagder:
Ok, "params" it is.

PR #100, opened 2023-04-06
make --json output query pairs under "params"

bagder:
A take at fixing #90, but I'm not sure of the syntax. Also, no handling of multiple identical keys yet.

### DIFF
--- a/test.pl
+++ b/test.pl
@@ -77,12 +77,24 @@ my %json_tests = (
             "port" => "443",
             "path" => "/",
             "query" => "utm=tra cker&address =home&here=now&thisthen",
-            "params" => {
-               "utm" => "tra cker",
-               "address " => "home",
-               "here" => "now",
-               "thisthen" => ""
-             }
+             "params" => [
+                 {
+                     "key" => "utm",
+                     "value" => "tra cker",
+                 },
+                 {
+                     "key" => "address ",
+                     "value" => "home",
+                 },
+                 {
+                     "key" => "here",
+                     "value" => "now"
+                 },
+                 {
+                     "key" => "thisthen",
+                    "value" => ""
+                 }
+             ]
         }
     ],
     "ftp://smith:secret\@example.com:33/path?search=me#where" => [
@@ -96,10 +108,12 @@ my %json_tests = (
             "password" => "secret",
             "query" => "search=me",
             "fragment" => "where",
-             "params" => {
-                 "search" => "me"
-            }
-            ,
+            "params" => [
+                {
+                    "key" => "search",
+                    "value" => "me"
+                }        
+            ]
         }
     ],
     "example.com" => [
--- a/test.pl
+++ b/test.pl
@@ -69,6 +69,22 @@ my @t = (
 );
 
 my %json_tests = (
+    "\"https://example.com?utm=tra%20cker&address%20=home&here=now&thisthen\"" => [
+        {
+            "url" => "https://example.com/?utm=tra%20cker&address%20=home&here=now&thisthen",
+            "scheme" => "https",
+            "host" => "example.com",
+            "port" => "443",
+            "path" => "/",
+            "query" => "utm=tra cker&address =home&here=now&thisthen",
+            "params" => {
+               "utm" => "tra cker",
+               "address " => "home",
+               "here" => "now",
+               "thisthen" => ""
+             }
+        }
+    ],
     "ftp://smith:secret\@example.com:33/path?search=me#where" => [
         {
             "url" => "ftp://smith:secret\@example.com:33/path?search=me#where",
--- a/test.pl
+++ b/test.pl
@@ -79,7 +79,11 @@ my %json_tests = (
             "user" => "smith",
             "password" => "secret",
             "query" => "search=me",
-            "fragment" => "where"
+            "fragment" => "where",
+             "params" => {
+                 "search" => "me"
+            }
+            ,
         }
     ],
     "example.com" => [
--- a/trurl.c
+++ b/trurl.c
@@ -529,7 +529,7 @@ static void json(struct option *o, CURLU *uh)
   }
   if(nqpairs) {
     int i;
-    fputs(",\n    \"q\": {\n", stdout);
+    fputs(",\n    \"params\": {\n", stdout);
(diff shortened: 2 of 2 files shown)


emanuele6:
URLs can technically contain two query entries with the same key.
`trurl --json` will output a URL with two query entries with the same keys as:
```bash
bash-5.1$ trurl --json 'example.org?x=1&v=3&x=2'
[
kk
  {
    "url": "http://example.org/?x=1&x=2",
    "scheme": "http",
    "host": "example.org",
    "port": "80",
    "path": "/",
    "query": "x=1&x=2",
    "q": {
      "x": "1",
      "v": "3",
      "x": "2"
    }
  }
]
```
Technically this is valid JSON, it is valid to have two entries in a JSON object with the same key, but most JSON parsing libraries/tools ignore this detail, and just keep one of the two values:
```
bash-5.1$ trurl --json 'example.org?q=1&v=3&q=2' | python -c 'import json, sys; print(json.dumps(json.load(sys.stdin)))'
[{"url": "http://example.org/?q=1&v=3&q=2", "scheme": "http", "host": "example.org", "port": "80", "path": "/", "query": "q=1&v=3&q=2", "q": {"q": "2", "v": "3"}}]
bash-5.1$ trurl --json 'example.org?x=1&v=3&x=2' | jq
[
  {
    "url": "http://example.org/?x=1&v=3&x=2",
    "scheme": "http",
    "host": "example.org",
    "port": "80",
    "path": "/",
    "query": "x=1&v=3&x=2",
    "q": {
      "x": "2",
      "v": "3"
    }
  }
]
bash-5.1$ trurl --json 'example.org?x=1&v=3&x=2' > foo.json; node -p 'JSON.stringify(require("./foo.json"))'
[{"url":"http://example.org/?q=1&v=3&q=2","scheme":"http","host":"example.org","port":"80","path":"/","query":"q=1&v=3&q=2","q":{"q":"2","v":"3"}}]
```

Maybe we could use a format like `"q": { "x": [ 1, 2 ], "y": 3 }` or `"q": [{"key":"x","value":"1"},{"key":"y","value":"3"},{"key":"x","value":"2"}` instead.

Also, even though, that JSON is technically valid, the JSON specification does not give meaning to the order of entries in a JSON object.

If the order of `v=2&x=10` vs `x=10&v=2` matters, then only the array of key-value pairs solution works.

I am not sure if this matters much.

emanuele6:
For what is worth, the javascript interface to parse query strings mentioned in #100 that is also available in `node`, returns a ordered collection of key value pair \(like `"params": [{"key":"x","value":"1"},{"key":"y","value":"3"},{"key":"x","value":"2"}]`\):
```bash
bash-5.1$ node -p 'new URLSearchParams(process.argv[1])' 'foo=1&bar=2&x&y&x&foo=3'
URLSearchParams { 'foo' => '1', 'bar' => '2', 'x' => '', 'y' => '', 'x' => '', 'foo' => '3' }
```

`python`'s builtin URL parsing library has returns a list of key-value tuples:
```bash
bash-5.1$ python -c 'import sys, urllib.parse; print(urllib.parse.parse_qsl(sys.argv[1], keep_blank_values=True))' 'foo=1&bar=2&x&y&x&foo=3'
[('foo', '1'), ('bar', '2'), ('x', ''), ('y', ''), ('x', ''), ('foo', '3')]
```
But it also provides an alternate convenience function that returns an unordered dictionary dictionary of keys to ordered list of values \(like `"params": { "x": [ 1, 2 ], "y": 3 }`\):
```bash
bash-5.1$ python -c 'import sys, urllib.parse; print(urllib.parse.parse_qs(sys.argv[1], keep_blank_values=True))' 'foo=1&bar=2&x&y&x&foo=3'
{'foo': ['1', '3'], 'bar': ['2'], 'x': ['', ''], 'y': ['']}
```

emanuele6:
I think the `URLSearchParams`/`urllib.parse.parse_qsl` way is the most correct output format.

So for the URL in the previous example, `trurl` would output something like:
```bash
bash-5.1$ trurl --json 'example.org?x=1&v=3&x=2'
[
  {
    "url": "http://example.org/?x=1&v=3&x=2",
    "scheme": "http",
    "host": "example.org",
    "port": "80",
    "path": "/",
    "query": "x=1&v=3&x=2",
    "params": [
      {
        "key": "x",
        "value": "1"
      },
      {
        "key": "v",
        "value": "3"
      },
      {
        "key": "x",
        "value": "2"
      }
    ]
  }
]
```

"key"/"value" could be shortened to "k"/"v", but I think "key"/"value" is better because it is compatible with `jq`'s `from_entries` function, so users can easily write a jq command like `jq '.[].params |= from_entries'` if they want to convert "params" to an object.
```bash
bash-5.1$ jq '.[].params |= from_entries' << 'EOF'
[
  {
    "url": "http://example.org/?x=1&v=3&x=2",
    "scheme": "http",
    "host": "example.org",
    "port": "80",
    "path": "/",
    "query": "x=1&v=3&x=2",
    "params": [
      {
        "k": "x",
        "v": "1"
      },
      {
        "k": "v",
        "v": "3"
      },
      {
        "k": "x",
        "v": "2"
      }
    ]
  }
]
EOF
jq: error (at <stdin>:24): Cannot use null (null) as object key
```
```bash
bash-5.1$ jq '.[].params |= from_entries' << 'EOF'
[
  {
    "url": "http://example.org/?x=1&v=3&x=2",
    "scheme": "http",
    "host": "example.org",
    "port": "80",
    "path": "/",
    "query": "x=1&v=3&x=2",
    "params": [
      {
        "key": "x",
        "value": "1"
      },
      {
        "key": "v",
        "value": "3"
      },
      {
        "key": "x",
        "value": "2"
      }
    ]
  }
]
EOF
[
  {
    "url": "http://example.org/?x=1&v=3&x=2",
    "scheme": "http",
    "host": "example.org",
    "port": "80",
    "path": "/",
    "query": "x=1&v=3&x=2",
    "params": {
      "x": "2",
      "v": "3"
    }
  }
]
```

bagder:
Switched to the params and key/value approach

bagder:
So maybe I should proceed and merge this and let everyone have a try and see how it runs?

Issue #88, opened 2023-04-06
Query string sort


cisene:
Feature:
* order query string variables according to supplied list
* sort query string variables ascending/descending

Order query string variables according to list:
`$ trurl "https://example.com?echo=foo&lima=foo&november=foo" -qsl '[november, lima, echo]' ` would produce
https://example.com?november=foo&lima=foo&echo=foo

Sorted ascending or descending:
`$ trurl "https://example.com?echo=foo&lima=foo&november=foo" -qsa ` would produce
https://example.com?echo=foo&lima=foo&november=foo

`$ trurl "https://example.com?echo=foo&lima=foo&november=foo" -qsd ` would produce
https://example.com?novemmber=foo&lima=foo&echo=foo

Background: I do a lot of URL manipulation to de-duplicate Podcast feed links (for PodcastIndex.org), by sorting query string variables or ordering them in a desired fashion, duplicates can be found and removed.

bagder:
I like this idea to enhance a user's ability to compare URLs using queries.

bagder:
Thinking more on this. Don't we cover most uses cases with just a single alphabetical sort? I mean, does it really add much usability with the lists thing and reversed alphabetical orders?

bagder:
In the mean time I merged support for `--sort-query` to allow users to try it out and see if there are actual remaining use cases for other sort methods that this single sort approach does not satisfy.

PR #101, opened 2023-04-06
add --sort-query

bagder:
Sorts the query pairs in a case insensitive alphabetical order

Added a test. Added to the man page.

Ref: #88


bagder:
This is a first shot at sorting. If we really want to do *other* sort alternatives as well, we might want to alter how this flag works/is used.

Issue #102, opened 2023-04-06
-- is not supported


emanuele6:
`trurl` doesn't support the standard `--` argument to indicate end of options

```
bash-5.1$ trurl --
trurl error: unknown option: --
trurl error: Try trurl -h for help
bash-5.1$ trurl -- hi
trurl error: unknown option: --
trurl error: Try trurl -h for help
```

bagder:
What are you proposing it should do?

emanuele6:
It should work like in other tools, it should be ignored, and arguments after it should not be treated as options even if they start with `-`.

Currenlty it is not possible to pass arguments that start with `-`, and also if a user blindly passes arguments to `trurl`, some of those arguments could be unintentionally interpreted as options:

```bash
#!/bin/bash --
urls=( example.org --json emanuele6:hunter2@github.com ) # got from somewhere, e.g. "$@"
trurl -s scheme=https "${urls[@]}"
```
Output:
```json
[
  {
    "url": "https://example.org/",
    "scheme": "https",
    "host": "example.org",
    "port": "443",
    "path": "/"
  },
  {
    "url": "https://emanuele6:hunter2@github.com/",
    "scheme": "https",
    "user": "emanuele6",
    "password": "hunter2",
    "host": "sdf.org",
    "port": "443",
    "path": "/"
  }
]
```
Instead of the expected:
```
https://example.org/
https://--json/
https://emanuele6:hunter2@github.com/
```

`trurl -s scheme=https -- "${urls[@]}"` would fix this problem.

PR #104, opened 2023-04-06
trurl: support "--" end-of-options argument

emanuele6:


emanuele6:
Fixes #102

bagder:
But what's the point? To allow the user to provide a URL that starts with a dash that won't work?

emanuele6:
@bagder they will work, you can see it with `-f -` or `--url`:
```bash
bash-5.1$ printf %s\\n --json | ./trurl -f -
http://--json/
bash-5.1$ trurl --url --json
http://--json/
```
The parser will guess it is a `http` URL.

But that is not main reason why I would like this; if I get multiple URLs to process from an unknow source, to process them with trurl,I will have to either loop over them, and construct a second list of arguments that has a `--url` argument for each one:

```bash
urls=( example.org -a query=destination=notme curl.se/trurl ) # hardoded for this example
trurl_args=()
for url in "${urls[@]}"; do
     trurl_args+=( --url "$url" )
done
trurl --json "${trurl_args[@]}"
```

Invoke trurl multiple times so I can easily use `--url`, and use another tool to join the output together (with this approach I lose the ability to easily use `--verify` or check the exit code of trurl without adding extra code):
```bash
urls=( example.org -a query=destination=notme curl.se/trurl ) # hardoded for this example
for url in "${urls[@]}"; do
    trurl --json --url "$url"
done |
jq -s 'add'
```

If I am 100% elements of `${urls[@]}` will not contain newline, and will not be empty, or I loop to the array to verify that is true and discard elements that contain newline or are empty, I can use:
```bash
urls=( example.org -a query=destination=notme curl.se/trurl ) # hardoded for this example
for i in "${!urls[@]}"; do
    case ${urls[i]} in
        ''|*$'\n'*) unset -v 'urls[i]' # or maybe abort
    esac
done
urls=( "${urls[@]}" )
printf %s\\n | trurl --json -f -
```

And finally, only if I am 100% elements of `${urls[@]}` will not start with "-", or I can loop the array to verify that that is true, and discard elements that start with "-" (assuming I don't want to keep them), I can use:
```bash
trurl --json "$@"
```

It would be much nicer if as user I would not have to do this verification, filtering or plumbing before calling `trurl`, and I could just use `--` safely:
```bash
trurl --json -- "$@"
```

bagder:
Well, you could just do `trurl --json --url "${urls[@]}"` to fix that issue, but okay. I can accept this change as it might make sense for some.

bagder (approved):

emanuele6:
Thank you!

To clarify `--url "${urls[@]}"` would only solve the problem for the first element of the array:

```
~/trurl $ urls=( --version example.org -a query=destination=notme curl.se/trurl )
~/trurl $ trurl "${urls[@]}"                                                trurl version 0.3 libcurl/7.88.0
~/trurl $ trurl --url "${urls[@]}"
http://--version/?destination=notme
http://example.org/?destination=notme
http://curl.se/trurl?destination=notme
~/trurl $ trurl -- "${urls[@]}"
http://--version/
http://example.org/
http://-a/
trurl note: Bad hostname [query=destination=notme]
http://curl.se/trurl
```

PR #107, opened 2023-04-08
Allow building with msvc again

krishean:
Commit b0e4b3c introduced the `strcasecmp` function, which msvc does not have, this maps it to one that does exist.

The other define for `strdup` is to silence a compiler warning about msvc wanting the ISO name for the function instead of the POSIX name (warning C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C++ conformant name: _strdup. See online help for details.)


krishean:
Alright I didn't know it would automatically add things I was working on over in the other repo here. I was getting things ready for a msbuild CI job, but I'm not sure of a nice way to get the latest curl dependencies.

bagder:
When you make a pull request, the PR will be updated when you update the branch. This PR seems to include a lot of extra things not related to the subject.

I think we should make a proper Windows build that works out of the box, but one thing per PR is a good policy.

bagder:
If you prefer me to, I can cherry-pick just the fix commit and then let you work on a follow-up PR for a windows build?

krishean:
Yeah, that seems like a good idea. I had originally intended to make a different request for the windows build files.

bagder:
Thanks, done!

PR #110, opened 2023-04-08
Add msvc build files

krishean:
This adds build files for vc2022, and the needed include files and static libraries built from curl 8.0.1.

A CI job should be able to be added from: https://github.com/actions/starter-workflows/blob/main/ci/msbuild.yml with very minor chages, the `run: msbuild` line included in the example yml probably doesn't need any changes.

Other example msbuild commands:
```
msbuild /m /t:Clean,Build /p:Configuration=Debug /p:Platform=x86 trurl.sln
msbuild /m /t:Clean,Build /p:Configuration=Release /p:Platform=x86 trurl.sln
msbuild /m /t:Clean,Build /p:Configuration=Debug /p:Platform=x64 trurl.sln
msbuild /m /t:Clean,Build /p:Configuration=Release /p:Platform=x64 trurl.sln
```

I'll have to add in arm64 support in a future request, but I don't have the platform tools for that setup just yet.

### DIFF
--- a/winbuild/README.md
+++ b/winbuild/README.md
@@ -1,0 +1,61 @@
+
+# Building trurl with Microsoft C++ Build Tools
+
+Download and install [Microsoft C++ Build Tools](https://visualstudio.microsoft.com/visual-cpp-build-tools/)
+
+When installing, choose the `Desktop development with C++` option.
+
+## Open a command prompt
+
+Open the **x64 Native Tools Command Prompt for VS 2022**, or if you are on an x86 platform **x86 Native Tools Command Prompt for VS 2022**
+
+## Set up the vcpkg repository
+
+Note: The location of the vcpkg repository does not necessarily need to correspond to the trurl directory, it can be set up anywhere. But it is recommended to use a short path such as `C:\src\vcpkg` or `C:\dev\vcpkg`, since otherwise you may run into path issues for some port build systems.
+
+Once you are in the console, run the below commands to clone the vcpkg repository and set up the curl dependencies:
+
+~~~
+git clone https://github.com/microsoft/vcpkg.git
+.\vcpkg\bootstrap-vcpkg.bat
+.\vcpkg\vcpkg install curl:x86-windows-static-md
+.\vcpkg\vcpkg install curl:x64-windows-static-md
+.\vcpkg\vcpkg integrate install
+~~~
+
+Once the vcpkg repository is set up you do not need to run these commands again. If a newer version of curl is released, you may need to run `git pull` in the vcpkg repository and then `vcpkg upgrade` to fetch the new version.
+
+## Build in the console
+
+Once the vcpkg repository and dependencies are set up, go to the winbuild directory in the trurl sources:
+
+    cd trurl\winbuild
+
+Then you can call the build command with the desired parameters. The builds will be placed in an output directory as described below.
+
+## Parameters
+
+ - The `Configuration` parameter can be set to either `Debug` or `Release`
+ - The `Platform` parameter can be set to either `x86` or `x64`
+
+## Build commands
+
+ - x64 Debug: `msbuild /m /t:Clean,Build /p:Configuration=Debug /p:Platform=x64 trurl.sln`
+ - x64 Release: `msbuild /m /t:Clean,Build /p:Configuration=Release /p:Platform=x64 trurl.sln`
+ - x86 Debug: `msbuild /m /t:Clean,Build /p:Configuration=Debug /p:Platform=x86 trurl.sln`
+ - x86 Release: `msbuild /m /t:Clean,Build /p:Configuration=Release /p:Platform=x86 trurl.sln`
+
+Note: If you are using the x64 Native Tools Command Prompt you can also run the x86 build commands.
+
+## Output directories
+
+The output files will be placed in: `winbuild\bin\<Platform>\<Configuration>\`
+PDB files will be generated in the same directory as the executable for Debug builds, but they will not be generated for release builds.
+
+Intermediate files will be placed in: `winbuild\obj\<Platform>\<Configuration>\`
+These include build logs and obj files.
+
+## Tests
+
+Tests can be run by going to the directory of the output files in the console and running `perl .\..\..\..\..\test.pl`
+You will need perl installed to run the tests, such as [Strawberry Perl](https://strawberryperl.com/)
--- a/winbuild/trurl.sln
+++ b/winbuild/trurl.sln
@@ -1,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.5.33516.290
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "trurl", "trurl.vcxproj", "{575657CF-843F-491C-B15B-881C28DF36CA}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
+		Release|x64 = Release|x64
+		Release|x86 = Release|x86
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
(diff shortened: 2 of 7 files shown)


bagder:
I don't think we should get a whole libcurl binary and include tree into our repository for this. A build instruction should detail how to get get and install that prior to a build.

krishean:
I wasn't sure of a good way to include the static libraries and include files, since the situation with curl over on nuget was somewhat of a mess. I have a batch script for building the curl static libraries that I can probably adapt for some step by step build instructions.

dfandrich:
There are also official (and unofficial) Windows curl binaries at https://curl.se/download.html#Win32

krishean:
I see the official builds all use mingw, and I had forgotten about vcpkg since the last time I used it, it generated several gigabytes of intermediate files for the half dozen static libraries I needed. I'll give vcpkg a try again for building the curl static libraries.

krishean:
Running the commands:
```
vcpkg install curl:x64-windows-static
vcpkg install curl:x86-windows-static
vcpkg export curl:x64-windows-static curl:x86-windows-static --raw
```
And then copying the `vcpkg-export-*\installed\x64-windows-static` and `vcpkg-export-*\installed\x86-windows-static` directories to the project directory should get it to build.

bagder:
> This branch cannot be rebased due to conflicts

Can you please rebase and force-push to fix these?

Also, can you please provide a build instruction for the documentation? For bonus points: use the build instructions to also make a new CI job that performs a build and runs the test suite.

krishean:
I don't see a way to rebase and force-push from the web ui, it seems quite limited. I can delete this branch (which would close this pr), and create a new pr with just the two files and one change.

I do see there are ways to use vcpkg in github CI, but it'a a little above anything I have experience with, a lot of github is new to me.

bagder:
> I don't see a way to rebase and force-push from the web ui

Then I strongly urge you to not rely on the GitHub's web UI.

> I do see there are ways to use vcpkg in github CI, but it'a a little above anything I have experience with, a lot of github is new to me.

Let's take one step at a time then. If we can get a working build with instructions merged, maybe someone else can help us out to get a CI job doing it. I personally don't know Windows much at all so I'm also not a suitable person to do it.

krishean:
I think that did it? It's an awful mess of commits though.

I'll have to write up a readme.md file for the build instructions.

krishean:
So after a long time of wondering why the project files looked wrong I finally figured out that if they are initially created with a version of visual studio that does not include support for vcpkg, but then are later opened with a newer version that does, the newer version will not add the necessary attributes to the project files. It simply acts as though vcpkg does not exist, you have to completely regenerate the project files with the newer version of visual studio to get the option to show up.

All that being said, I _think_ this should build with CI if a job is set up for it. If anyone has a windows machine handy to verify the instructions in the readme work, or has any suggestions for improving the readme I would appreciate it.

winbuild/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Note: The location of the vcpkg repository does not necessarily need to correspond to the trurl directory, it can be set up anywhere. But it is recommended to use a short path such as `C:\src\vcpkg` or `C:\dev\vcpkg`, since otherwise you may run into path issues for some port build systems.
+
+Once you are in the console, run the below commands to clone the vcpkg repository and set up the curl dependencies:
+
+~~~
+git clone https://github.com/microsoft/vcpkg.git
+.\vcpkg\bootstrap-vcpkg.bat
+.\vcpkg\vcpkg install curl:x86-windows-static-md
+.\vcpkg\vcpkg install curl:x64-windows-static-md
```
bagder, line 22: Why do we need to install two versions? Isn't the x86 one for legacy 32 bit builds? How about just dropping that and pretending they don't exist? I seems unlikely that the rare remaining 32 bit  users could not figure out how to fix this if they need to, and leave those out from current users on 64 bit systems which probably is 99% of the population.

winbuild/trurl.sln
```diff
@@ -0,0 +1,31 @@
...
+		{575657CF-843F-491C-B15B-881C28DF36CA}.Release|x86.ActiveCfg = Release|Win32
+		{575657CF-843F-491C-B15B-881C28DF36CA}.Release|x86.Build.0 = Release|Win32
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {14A4D782-313F-4F61-A2C5-EF2CD877D3F3}
+	EndGlobalSection
+EndGlobal
```
bagder, line 31: I presume this is generated somehow? Is this specific for this particular visual studio version? What if you run an older or a newer version?

Why is there both a `.sln` and a `.vcxproj` file? Are they both specific to this MSVC version?

bagder (commented):
I'm scared of the generated(?) files as they look version-specific (and proprietary). Will that mean that we need to provide those for every msvc version ever created? Can we leave them out from the repo and have the user create them themselves instead?

winbuild/README.md
krishean: I just included it for backwards compatibility, I really only use the x64 builds. I could change the readme to say you can install one, the other, or both, or if you would prefer remove it altogther. That part of the readme will need a small edit in the future for arm64 compatibility, when I get that build system working.

winbuild/trurl.sln
krishean: The `.sln` and `.vcxproj` files were initially generated by vs2019, then I edited the vcxproj file by hand a few times. I had to regenerate them with vs2022 because of a problem getting vcpkg working. Usually visual studio projects have both files, ~~but I can check if msbuild will compile using just the vcxproj file, without the sln file.~~ it looks like yes, it will compile from just the vcxproj file.

As far as I know the sln file does not have any version specific configuration in it, I was able to use the vs2019 generated one on vs2022 without it having any complaints. The vcxproj does have one version specific setting that I know of in it: `PlatformToolset`, but it can easily be changed to any of the values listed in the documentation here: https://learn.microsoft.com/en-us/cpp/build/how-to-modify-the-target-framework-and-platform-toolset?view=msvc-170#platform-toolset I could include a section in the readme about changing that for different visual studio versions.

bagder:
It probably helps for review if you squash all these commits into the lowest number of commits you think it should be and force-push.

I don't know anything about Windows, I need someone else's review on this and/or a CI job that verifies it.

krishean:
I've also been working on a set of cmake files for visual studio that may be a better option, as they can potentially be built cross platform. So you may want to hold off on merging this if you would prefer cmake files over the msbuild version.

jay:
Why is vcpkg needed at all for this? Wouldn't it be better if the user just links to their libcurl.lib or libcurld.lib or whatever

krishean:
I added the vcpkg config so we could have windows CI builds without including binaries in the repo, if there's another way to do that I'm not set on doing it that way

jay:
I would think that for CI purposes the latest curl would be cloned from the repo and built and that winbuild not contain any of that code. The [curl project has a winbuild](https://github.com/curl/curl/tree/master/winbuild) where the user can specify dependencies like SSL_PATH=..., so I would think the trurl project would have a winbuild makefile where the user can specify the libcurl dependency like LIBCURL_PATH=...

curl winbuild is tested using appveyor ci jobs, for examples see https://github.com/curl/curl/blob/master/appveyor.yml#L314 , however I notice there are no curl dependencies built from git with appveyor so github actions may be a better choice to clone and build curl, for examples see https://github.com/curl/curl/tree/master/.github/workflows

winbuild/trurl.sln
jay: There's a DefaultPlatformToolset variable, see https://www.appveyor.com/docs/lang/cpp/#platformtoolset and https://stackoverflow.com/questions/45474857/why-does-appveyor-use-vs2010-tools-for-vs2015-image

krishean:
That appveyor.yml is very interesting, I hadn't noticed it before (I didn't know that was what was used for the windows builds.) I see it does have a CMake section in there as well

I see in the linux.yml file it is cloning other repos to build dependencies, so I could take a look at that a bit more

krishean:
This successfully ran: https://github.com/krishean/trurl/blob/msvc-cmake/.github/workflows/winbuild.yml

PR #112, opened 2023-04-08
add support for building with older libcurl versions

bagder:
This still fails some tests due to missing functionality in the older versions.


bagder:
This is an alternative take to #85. I tested it with libcurl/7.74.0.

PR #114, opened 2023-04-08
Properly free memory from libcurl functions

dfandrich:
These pointers must be freed with curl_free, not free. Also, free some
memory that leaked otherwise. The test suite now passes with valgrind.

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -225,7 +225,7 @@ static void pathadd(struct option *o, const char *path)
   struct curl_slist *n;
   char *urle = curl_easy_escape(NULL, path, 0);
   if(urle) {
-    n = curl_slist_append(o->append_path, strdup(urle));
+    n = curl_slist_append(o->append_path, urle);
     if(n) {
       o->append_path = n;
     }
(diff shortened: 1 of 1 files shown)


trurl.c
```diff
@@ -225,11 +225,11 @@ static void pathadd(struct option *o, const char *path)
   struct curl_slist *n;
   char *urle = curl_easy_escape(NULL, path, 0);
   if(urle) {
-    n = curl_slist_append(o->append_path, urle);
+    n = curl_slist_append(o->append_path, strdup(urle));
```
jay, line 228: this expects valid param !=NULL so I think an additional check is needed, eg
char *dupe = strdup(urle);
if(!dupe)
....

jay: actually on second thought curl_slist_append should duplicate the data so actually I'm not sure what you're doing here

bagder: Yes, `curl_slist_append()` strdups the data it self so this should not be necessary!

bagder: This causes a leak that we can prove by running:

~~~
valgrind --leak-check=full ./trurl curl.se --append path=moo

bagder:
Thanks!

PR #115, opened 2023-04-08
CI: Add a UBSAN build

dfandrich:
The code doesn't currently pass the address or memory sanitizers, but
that should be a goal.


bagder:
Awesome!

PR #116, opened 2023-04-09
Fixed a few small memory leaks detected with valgrind.

jacobmealey:


jacobmealey:
Im a fool! this is already fixed in #114

Issue #40, opened 2023-04-03
Loop over multiple --sets


jacobmealey:
I'm not sure what it would look like but it would be helpful to generate a range of URLs for something like the port. I was thinking it could have a syntax similar to below but there is most certainly a better way! 
`./trurl --url google.com --set port=[1000-1005]`

or if it's something that would be a string instead of numbers it could take a list like:
`./trurl --url google.com --set host=[google.co.uk, google.co.nz]`

bagder:
Maybe we could have a separate option for that. Like `--iterate`  or something that can accept ranges and lists like proposed here.

jacobmealey:
That makes sense, and would probably require less rewriting of existing stuff. Were you thinking it would work something like: 

`trurl --url something.com --iterate hosts=[another.com, something.co.uk]` ? 

It probably wouldn't need the brackets, not sure why I am so fixated on them. It could probably be something like the following, but that may make parsing a little trickier:
`trurl --url something.com --iterate hosts=another.com, something.co.uk ports=80,443,8080`

bagder:
Something like that, yes. It might make more sense to use spaces as separators rather than comma though, simply because space cannot legally be part of any component.

And maybe one "loop" per `--iterate` option?

~~~
$  trurl --url something.com --iterate "host=another.com something.co.uk" --iterate "port=80 443 8080"

jacobmealey:
Excellent. I can start working on it this evening.

Earnestly:
Would it be unreasonable to just issue the command multiple times instead?

Earnestly:
That's fair.  Another possible design would be similar to curl's `-K config` which accepts a `key=value` format much like `-s, --set` does here.  E.g.

```
url=something.com host=another.com port=80
url=something.com host=another.com port=443
url=something.com host=another.com port=8080
url=something.com host=something.co.uk port=80
url=something.com host=something.co.uk port=443
url=something.com host=something.co.uk port=8080
```

Which can be generated with something like:

```
printf '%s\n' 'url=something.com host='{another.com,something.co.uk}' port='{80,443,8080}
```

(Although this use of brace expansion isn't POSIX sh yet: https://austingroupbugs.net/view.php?id=1193#c3995)

jacobmealey:
Folks, I just wanted to share a quick update. I have parsing of the `--iterate` flag working and it generates urls if only a single variable is being iterated. for example `trurl --url something.com --iterate "ports=80 443 8080"` produces:
```
[...]

PR #117, opened 2023-04-09
Trurl iterate sets

jacobmealey:
Folks, This PR is more of a call for help. 

As I mentioned in the issue ( #40) I have gotten it working for a single iterate, so I can loop over all the iterations for a particular value. What I am struggling with is making it work for many values. 

My thoughts were that `struct option` could be turned into a linked list, so each permutation of the url (each iteration specified by the user) may be represented as a single `option`. This provides minimal alterations to the existing code. 

I've cleaned it up so it works nicely for single iterations, so if you (anybody at all) can see how it would be extracted out that would be great.

OR if you think theres a better _high level_ approach to take the I would be very interested to hear.  

(also a note this will all be squashed and cleaned up before a final commit)


bagder:
I think you can just keep a list of `iterate_list` nodes in the struct option., I don't think you need to do many struct options. It doesn't change per iteration. It is the URL that is changed.

bagder:
Also, please stick to the existing code style: 2-space indents and the brace on column 0 when we define new functions.

jacobmealey:
Hi, thank you. I will be sure to make sure all the code is up to snuff for the final PR.

jacobmealey:
Alright, folks! sorry this took me so long, things have been busy on my end with the school semester coming to an end. 
I am pretty happy with how it shook out in the end. It says its failing tests, but the tests are failing on master as well so not sure what I can do about that.

bagder:
The tests are not failing on master. The most recent merge is all green and they run fine for me locally.

PR #121, opened 2023-04-09
fix: small styling issue

sfmqrb:


trurl.c
```diff
@@ -60,7 +60,7 @@
 #define OUTPUT_QUERY    8
 #define OUTPUT_FRAGMENT 9
 #define OUTPUT_ZONEID   10
-#define NUM_COMPONENTS 11 /* including "url" */
+#define NUM_COMPONENTS  11 /* including "url" */
```
bagder, line 63: I don't think this is necessary, these are different lists and we should rather separate them with a newline or something.

bagder:
That line is now separated from the list with a newline, I don't think they need to be aligned.

PR #123, opened 2023-04-10
checksrc: check code style in CI job and with 'make checksrc'

bagder:
The checksrc script is imported from the curl source repo (updated last as 2bc1d775f51)


bagder:
Experience says code style is best kept in order with a tool.

PR #131, opened 2023-04-13
add install instruction with tea

mfts:
We saw that currently there's no package manager to support `trurl`, so we wanted to change that 😉

[`tea`](https://github.com/teaxyz/cli) is a new kind of package manager from the creator of Homebrew and we packaged `trurl` up so that anyone using tea can easily install and use `trurl` as well :)


mfts:
> I think https://github.com/curl/trurl/discussions/51 tells a different story.

My bad. 

> Maybe a separate INSTALL.md which could include details for everyone that provides the info?

I'll add it to #51 and close this PR for now. I'll add an INSTALL.md next week.

Issue #129, opened 2023-04-11
Issues with --get on windows powershell


havbon:
There seems to be some issues when using --get on windows, here is the log from test.py
[testout.log](https://github.com/curl/trurl/files/11201260/testout.log)

When executing `--get {url}` it outputs `url`

bagder:
That's because it's not quoted isn't it?

havbon:
![image](https://user-images.githubusercontent.com/46850782/231225781-1e960611-b573-4d24-b58a-aa427f45bc3c.png)

i'm afraid not, ill do some more testing in a bit

havbon:
a few more examples:
![image](https://user-images.githubusercontent.com/46850782/231243640-344509e9-5339-4fc7-81c4-9f6f8ec4703b.png)


It probably has something to do with:

> To create or display a variable name that includes spaces or special characters, enclose the variable name with the curly braces ({}) characters. The curly braces direct PowerShell to interpret the variable name's characters as literals.

from [powershell docs](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_variables?view=powershell-7.3)

But it's difficult to tell...

emanuele6:
I can't reproduce this on powershell on linux:
```
PowerShell 7.3.3
PS /home/emanuele6> trurl localhost -g {host}
trurl error: unknown option: -inputFormat
trurl error: Try trurl -h for help
PS /home/emanuele6> trurl localhost -g '{host}'
localhost
PS /home/emanuele6> trurl localhost -g "{host}"
localhost
PS /home/emanuele6>
```

`{host}` does not expand if you use quote it either using `"` or `'`. It only expands when not quoted as described by the documentation.

<https://learn.microsoft.com/en-us/cpp/cpp/main-function-command-line-args?view=msvc-170>
IIRC, on Windows, processes are created with one single argument, not with a vector or multiple arguments (`argv`) like in UNIX systems.  And how the quoting is interepreted/wildcards/etc of the single argument are interpreted is up to the individual process; in the case of C programs with a main() function, how the quoting is handled is decided by the startup code that calls main added by the linker/compiler.

Maybe check the documentation of the build tools you used to compile to `trurl` on Windows to see if you can find how `{host}` should be quoted, or any other information about this.

emanuele6:
Maybe your trurl.exe understands that `-encodedCommand` stuff.
[...]

PR #132, opened 2023-04-14
trurl: make '{scheme}\{user}' work

emanuele6:
trurl: support \\\\ in -g formats

This allows e.g. printing 'imap\\new' with  -g '{host}\\\\new'.
It was previously not possible to print a literal backslash followed by
a n/t/r with -g.

---

trurl: make '{scheme}\\{user}' work

Before this patch, the sequence \`\\{' was handled by printing a literal
backslash followed by a literal {.

This was a bit awkward since you are supposed to be using \`{{' to escape
a { character, not \\. Also because the backslash was still being
printed.

With this patch \`\\{' is now handled by only printing a backslash, and
then handling the next character, instead of printing it
indiscrimintately.

So, now,   trurl -g '{scheme}\\{user}' gopher://emanuele6@locahost/
outputs   gopher\\emanuele6   instead of   gopher\\{user}   .

The previous commit that added support for \`\\\\' makes this change less
necessary since the user should have used   {scheme}\\\\{user}   instead
of using a single backslash, but I still think this behaviour makes more
sense, and may prevent future bugs.

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -424,6 +424,7 @@ static void get(struct option *op, CURLU *uh)
         ptr++; /* pass the { */
         if(!end) {
           /* syntax error */
+          fputc('{', stream);
           continue;
         }
         /* {path} {:path} */
--- a/trurl.c
+++ b/trurl.c
@@ -494,10 +494,10 @@ static void get(struct option *op, CURLU *uh)
         fputc('\\', stream);
         break;
       default:
-        /* unknown, just output this */
+        /* unknown, just output the backslash, and continue */
         fputc(*ptr, stream);
-        fputc(ptr[1], stream);
-        break;
+        ptr++;
+        continue;
       }
       ptr += 2;
     }
--- a/trurl.c
+++ b/trurl.c
@@ -490,6 +490,9 @@ static void get(struct option *op, CURLU *uh)
       case 't':
         fputc('\t', stream);
         break;
+      case '\\':
+        fputc('\\', stream);
+        break;
       default:
         /* unknown, just output this */
         fputc(*ptr, stream);


emanuele6:
fixup: changed `++ptr` to `ptr++` for consistency with the rest of the code.

emanuele6:
I learned about those backslash sequence, and `{{` only after reading the source code, because I noticed that `-g '\{scheme}'` printed `\{scheme}` literally when playing with powershell for #129.
I didn't know about them because they are not documented.

The addition of `\\` with the first commit, really makes the second commit unnecessary since it would only change the behaviour of `\unknown` which, I assume, at least once those backslash sequences will be documented, will be unspecified.

The current way `trurl` handles unknown sequences is printing both the backslash character and the following character literally, and then handling the second character after the backslash. This way `\s` for example, is still printed as `\s` as if `\\s` was used.

A side effect of the current implementation is that `\{host}` is equivalent to `\\{{host}`, since the code skips the `{` following the backslash and prints it indiscriminately, which I find weird. So, I changed the behaviour of `\unknown` to make it equivalent to `\\unknown` which was probably the behaviour that was originally intended.

bagder:
I think the backslash handling should work like the `printf` command line tool. Not introduce a new way of working. For \-[unknown symbol] it outputs both the backslash and the symbol:

~~~
$ printf '\\ hello \{\n'
\ hello \{
~~~

where as with this PR, trurl hides the unknown symbol:
~~~
$ ./trurl localhost -g '{host} \\ hej \{ moo \'
localhost \ hej \ moo \
~~~

emanuele6:
> where as with this PR, trurl hides the unknown symbol:
>
> ```
> $ ./trurl localhost -g '{host} \\ hej \{ moo \'
> localhost \ hej \ moo \
> ```

That is not supposed to happen; I think this is an existing bug with unterminated `{component}` expansions:
```
$ ./trurl localhost -g '{foo'
foo
```

emanuele6:
Closing in favour of #133

PR #133, opened 2023-04-14
support --get variables within [brackets]

bagder:
also extend the backslash support somewhat and document it.

### DIFF
--- a/trurl.1
+++ b/trurl.1
@@ -48,11 +48,13 @@ Each line needs to be a single valid URL and the line needs to end with a
 newline. The maximum URL length supported in a file like this is 4095 bytes.
 .IP "-g, --get [format]"
 Output URL data according to the provided format string. Components from the
-URL can be output when specified as \fB{component}\fP, with the name of the
-part show within curly braces. The following component names are available
-(case sensitive): {url}, {scheme}, {user}, {password}, {options}, {host},
-{port}, {path}, {query}, {fragment} and {zoneid}. All other text in the format
-string will be shown as-is.
+URL can be output when specified as \fB{component}\fP or \fB[component]\fP,
+with the name of the part show within curly braces or brackets. You can not
+mix braces and brackets for this purpose in the same command line.
+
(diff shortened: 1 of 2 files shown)


bagder:
It's rather *additionally*

trurl.c
```diff
@@ -420,7 +429,7 @@ static void get(struct option *op, CURLU *uh)
         size_t vlen;
         int i;
         bool urldecode = true;
-        end = strchr(ptr, '}');
+        end = strchr(ptr, endbyte);
         ptr++; /* pass the { */
         if(!end) {
           /* syntax error */
```
emanuele6, line 435:
```suggestion
        if(!end) {
          /* syntax error */
          fputc(startbyte, stream);
```
To make `trurl -g '{scheme} {host' https://localhost` output `https {host` instead of `https host` as mentioned in #132

bagder:
Thanks, we obviously need to add a bunch of test cases to verify this.

trurl.1
```diff
@@ -48,11 +48,13 @@ Each line needs to be a single valid URL and the line needs to end with a
...
-(case sensitive): {url}, {scheme}, {user}, {password}, {options}, {host},
-{port}, {path}, {query}, {fragment} and {zoneid}. All other text in the format
-string will be shown as-is.
+URL can be output when specified as \fB{component}\fP or \fB[component]\fP,
+with the name of the part show within curly braces or brackets. You can not
+mix braces and brackets for this purpose in the same command line.
+
+The following component names are available (case sensitive): url, scheme,
+user, password, options, host, port, path, query, fragment and zoneid. All
+other text in the format string will be shown as-is.
```
emanuele6, line 57: Maybe "All other text in the format string will be shown as-is." should be moved to the very bottom, after the backslash sequences are also defined, or removed.
Here it may be interpreted as saying that `{somethingelse}` is shown as literally `{somethingelse}` which is not actually what happens.

bagder: Good idea. Attempt pushed.

PR #135, opened 2023-04-14
Fix test.pl on windows

krishean:
Windows doesn't like single quotes in commands, it causes those tests to fail (the others still pass.)

Changing single quotes to escaped double quotes makes all the tests pass again.

### DIFF
--- a/test.pl
+++ b/test.pl
@@ -88,7 +88,7 @@ my @t = (
     "localhost -g \"[scheme] [host\"|http [host",
     # two backslashes in the source end up one in the actual command line
     "localhost -g \"\\{{scheme}\\[\"|{http[",
-    "localhost -g \"\\\\\[\"|\\[",
+    "localhost -g " . (($^O eq 'MSWin32')?"\"\\\\\[\"":"'\\\\\['") . "|\\[",
     "https://u:s\@foo?moo -g \"[scheme][user][password][query]\"|httpsusmoo",
 );
 
--- a/test.pl
+++ b/test.pl
@@ -169,21 +169,23 @@ my %json_tests = (
 
 plan tests => keys(@t) + keys(%json_tests);
 
+my ( $cmd_string, $null_device ) = ($^O eq 'MSWin32')?(".\\trurl.exe", "nul"):("./trurl", "/dev/null");
+
 for my $c (@t) {
     my ($i, $o) = split(/\|/, $c);
     # A future version should also check stderr
-    my @out = ($^O eq 'MSWin32')?`.\\trurl.exe $i 2>nul`:`./trurl $i 2>/dev/null`;
+    my @out = `$cmd_string $i 2>$null_device`;
     my $result = join("", @out);
     chomp $result;
-    is( $result, $o, "./trurl $i" );
+    is( $result, $o, "$cmd_string $i" );
 }
 
 while (my($i, $o) = each %json_tests) {
-    my @out = ($^O eq 'MSWin32')?`.\\trurl.exe --json $i 2>nul`:`./trurl --json $i 2>/dev/null`;
+    my @out = `$cmd_string --json $i 2>$null_device`;
     my $result_json = join("", @out);
     my $result = decode_json($result_json);
 
-    is_deeply($result, $o, "./trurl --json $i");
+    is_deeply($result, $o, "$cmd_string --json $i");
 }
 
 done_testing();
--- a/test.pl
+++ b/test.pl
@@ -87,7 +87,7 @@ my @t = (
     "localhost -g \"{scheme} {host\"|http {host",
     "localhost -g \"[scheme] [host\"|http [host",
     # two backslashes in the source end up one in the actual command line
-    "localhost -g \"\\{{scheme}\\[\"|{http[",
+    "localhost -g " . (($^O eq 'MSWin32')?"\"\\{{scheme}\\[\"":"'\\{{scheme}\\['") . "|{http[",
     "localhost -g " . (($^O eq 'MSWin32')?"\"\\\\\[\"":"'\\\\\['") . "|\\[",
     "https://u:s\@foo?moo -g \"[scheme][user][password][query]\"|httpsusmoo",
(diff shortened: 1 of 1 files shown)


krishean:
Alright why is test 80 failing on linux and cygwin, that backslash is clearly escaped?

krishean:
windows:
```
>.\trurl.exe localhost -g "\\["
\[

>.\trurl.exe localhost -g "\\\\["
\\[

>.\trurl.exe localhost -g '\\['
'\['
```
linux:
```
$ ./trurl localhost -g "\\["
[
$ ./trurl localhost -g "\\\\["
\[
$ ./trurl localhost -g '\\['
\[
```

why do I try to write software again

test.pl
```diff
@@ -84,12 +84,12 @@
...
-    "localhost -g '{scheme} {host'|http {host",
-    "localhost -g '[scheme] [host'|http [host",
+    "localhost -g \"{scheme} {host\"|http {host",
+    "localhost -g \"[scheme] [host\"|http [host",
     # two backslashes in the source end up one in the actual command line
-    "localhost -g '\\{{scheme}\\['|{http[",
-    "localhost -g '\\\\\['|\\[",
-    "https://u:s\@foo?moo -g '[scheme][user][password][query]'|httpsusmoo",
+    "localhost -g \"\\{{scheme}\\[\"|{http[",
+    "localhost -g \"\\\\\[\"|\\[",
```
emanuele6, line 91: In UNIX shells, `"\\"` is equivalent to a single backslash, while `'\\'` is two backslashes

krishean: Yeah, I know that the perl code is double escaped, it's just when it hits the shell the behavior is different (and this is frustrating.)

krishean:
Here's something even more annoying, under cmd `"\\["` and `'\\['` have different output, but under powershell it is the same:

powershell:
```
PS> .\trurl.exe localhost -g "\\["
\[
PS> .\trurl.exe localhost -g '\\['
\[
```

krishean:
Would it be better to use `#!/usr/bin/env perl` for the shebang?

emanuele6:
I don't know any platform that requires that.
In any case, `make test` runs the script with `perl test.pl` so it should not matter.

krishean:
As far as I know it's for unix like systems if you chmod +x the script so you can invoke it like: `./test.pl`, on windows the line is pretty much ignored entirely.

Sometimes I see comments that `#!/usr/bin/env [perl/bash/sh/python/etc]` is better for backwards compatibility, but other comments say it could be a security issue if someone manipulates the environment and the script is being invoked with privileges. So I'm not sure which would be better.

krishean:
Alright, this has been bothering me, when running under bash, should the output from `"\\["` be just `[` or did I stumble across some other bug?

bagder:
On bash:
~~~
$ ./trurl localhost -g "\\["
[
$ ./trurl localhost -g '\\['
\[

bagder:
But this is not bash specific, this will work like this on all/most unix shells.

krishean:
Both cmd and powershell have different results, and `echo "\\["` does output `\[` as well on bash.

krishean:
Would that also mean that test 79 would need to be treated as a special case, as it would not be passing the backslashes into trurl with enclosing double quotes on unix like shells?

bagder:
If the shell used for tests does not work like a POSIX shell, then that shell might not evaluate the command line the same way. My gut feeling says this will be the case for multiple test cases running on windows. So yeah, we should probably figure out a way to handle that.

krishean:
Test 79 passes on both windows and linux/cygwin as it is right now, but possibly for the wrong reasons. It would be getting the single backslash in the input on windows, but it would be omitted (treated as an escape) on linux and cygwin.

I'll put in another commit for that.

krishean:
I just noticed the python test was added, and uses a json file, would you want me to work on making the perl version point at that json file as well? It looks like it might also have the single quote issues on windows, which I could look at at the same time.

Or would the perl tests be replaced with the python version?

bagder:
> Or would the perl tests be replaced with the python version?

Probably, yes. But ultimately we just need a lot of tests and we need the tests to run on as many platforms as possible. I don't really care which language the test runner is written in, I just need it to work and do its job.

test.pl
```diff
@@ -84,12 +84,12 @@
...
-    "localhost -g '[scheme] [host'|http [host",
+    "localhost -g \"{scheme} {host\"|http {host",
+    "localhost -g \"[scheme] [host\"|http [host",
     # two backslashes in the source end up one in the actual command line
-    "localhost -g '\\{{scheme}\\['|{http[",
-    "localhost -g '\\\\\['|\\[",
-    "https://u:s\@foo?moo -g '[scheme][user][password][query]'|httpsusmoo",
+    "localhost -g " . (($^O eq 'MSWin32')?"\"\\{{scheme}\\[\"":"'\\{{scheme}\\['") . "|{http[",
+    "localhost -g " . (($^O eq 'MSWin32')?"\"\\\\\[\"":"'\\\\\['") . "|\\[",
+    "https://u:s\@foo?moo -g \"[scheme][user][password][query]\"|httpsusmoo",
```
bagder, line 92: Whatever adaptions is necessary for Windows, or any other platform, they should be done in a generic fashion so that the test cases can be written and used independently. Because we want to be able to test using single-quotes. If we cannot run-time translate those to working Windows alternatives, we need to detect that and rather skip the test on Windows. Converting every single-quote test to a double-quote on all platform is not a good idea.

krishean: I did the work last night to make the script read the json file rather than having all the tests hardcoded, so all of what I did here is pretty much irrelevant. I used shellwords from Text::ParseWords to parse the parameters before feeding them into start from IPC::Run as an array, which works on both linux and windows without altering any of the parameters in the json file.

Should we abandon this request and open a new one for the new set of changes, or just add them as an additional commit here?

bagder:
We will probably remove the perl version and stick with `test.py` going forward.

PR #141, opened 2023-04-16
bmake-compatible makefile

nekobbbbbbit:
Uses a BSD extension (!=) which GNU Make seems to support. Allows me to remove the gmake dependency for the OpenBSD port (I could use a patch here, but I didn't see how this could hurt in upstream)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ MANDIR ?= $(DESTDIR)$(PREFIX)/share/man/man1
 INSTALL = install
 
 $(TARGET): $(OBJS)
-	$(CC) $(CFLAGS) $(OBJS) -o $(TARGET) $(LDLIBS)
+	$(CC) $(OBJS) -o $(TARGET) $(LDLIBS)
 
 trurl.o:trurl.c version.h
 
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 TARGET = trurl
 OBJS = trurl.o
 LDLIBS != curl-config --libs
-CFLAGS := $(CFLAGS) -W -Wall -pedantic -g
+CFLAGS != curl-config --cflags
+CFLAGS += -W -Wall -pedantic -g
 MANUAL = trurl.1
 
 PREFIX ?= /usr/local
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ MANDIR ?= $(DESTDIR)$(PREFIX)/share/man/man1
 INSTALL = install
 
 $(TARGET): $(OBJS)
-	$(CC) $(OBJS) -o $(TARGET) $(LDLIBS)
+	$(CC) $(OBJS) -o $(TARGET) $(LDLIBS) $(LDFLAGS)
 
 trurl.o:trurl.c version.h
 
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 TARGET = trurl
 OBJS = trurl.o
-LDLIBS = -lcurl
+LDLIBS != curl-config --libs
 CFLAGS := $(CFLAGS) -W -Wall -pedantic -g
 MANUAL = trurl.1
 
@@ -11,6 +11,7 @@ MANDIR ?= $(DESTDIR)$(PREFIX)/share/man/man1
 INSTALL = install
 
 $(TARGET): $(OBJS)
+	$(CC) $(CFLAGS) $(OBJS) -o $(TARGET) $(LDLIBS)
 
 trurl.o:trurl.c version.h
 


Makefile
```diff
@@ -1,6 +1,6 @@
 TARGET = trurl
 OBJS = trurl.o
-LDLIBS = -lcurl
+LDLIBS != curl-config --libs
```
bagder, line 3: Do you really need curl-config ? I would prefer sticking to plain old lib only if we can.

nekobbbbbbit: > Do you really need curl-config ? I would prefer sticking to plain old lib only if we can.

I need it in my case. curl-config adds the include header location. I could also not include it (do += for ldlibs, which seems required on OpenBSD and FreeBSD to see /usr/local/include), but I don't particularly see what's wrong with using Curl's -config tool as this does this already (and possibly other things).

emanuele6:
Personally, I also tried to build trurl on my sdf.org shell that runs NetBSD some days ago, and I noticed that it did not build out of the box because `-lcurl` is not sufficient:
```bash
$ curl-config --libs
-Wl,-zrelro -L/sys/pkg93/lib -Wl,-R/sys/pkg93/lib -L/usr/lib -Wl,-R/usr/lib -L/usr/lib -L/sys/pkg93/lib -L/sys/pkg93/lib -L/sys/pkg93/lib -L/sys/pkg93/lib -lcurl
```
But your patch is also not sufficient to fix the build on sdf.org, because, before even noticing that, i also noticed that i needed to add the output of `curl-config --cflags` to `CFLAGS`:
```bash
$ curl-config --cflags
-I/sys/pkg93/include
```
Otherwise the compiler was not able to find `<curl/curl.h>`.

---
Also, both using `gmake` and using `bmake`, I don't need to redefine the default `$(TARGET): $(OBJS)` rule. Are you sure you really need that?

Makefile
emanuele6: I think just `LDLIBS := $(LDLIBS) -lcurl` probably makes more sense, since in the patch you are only using `curl-config` for `--libs`, and not for `--cflags`.
Alternatively, you could use `curl-config` for both.

nekobbbbbbit:
> Also, both using `gmake` and using `bmake`, I don't need to redefine the default `$(TARGET): $(OBJS)` rule. Are you sure you really need that?

For some reason I was only getting trurl.o but not the target until I added that. Maybe I am wrong...

But good catch on the CFLAGS bit. Mainly because I have had to deal with so many broken CFLAGS forgetting my custom include that I just added an ENV cflag for it, lol.

emanuele6:
> > Also, both using `gmake` and using `bmake`, I don't need to redefine the default `$(TARGET): $(OBJS)` rule. Are you sure you really need that?
>
> For some reason I was only getting trurl.o but not the target until I added that. ~~Maybe I am wrong...~~, nope, I added that for a reason

I have not tried reproducing this on OpenBSD, but maybe that is only required on OpenBSD.

I looked at the online documentation for OpenBSD's `make` utility, and it doesn't seem to be standard BSD make: it's either an older version, or OpenBSD's own fork of BSD make; e.g. OpenBSD make doesn't have the `-T` option, but BSD make has it.

* [OpenBSD's `make(1)`](https://man.openbsd.org/make.1)
* [FreeBSD's `make(1)`](https://man.freebsd.org/cgi/man.cgi?make)
* [NetBSD's `make(1)`](https://man.netbsd.org/make.1)
* [DragonflyBSD's `bmake(1)`](https://man.dragonflybsd.org/?command=bmake)

We could add it for OpenBSD compatibility if it's really necessary I guess.

emanuele6:
Oh, I think I know what is happening.

BSD make doesn't have a builtin `.o` rule like GNU make, so `trurl: trurl.o` doesn't work, but it fallbacks to using the POSIX-mandated `.c` rule to build `trurl` from `trurl.c` directly, since it doesn't know how to build `trurl` with `trurl.o`.

I just noticed that the second command in `make`'s output that is supposed to be the one that links `trurl.o` actually uses `trurl.c`:
```bash
$ make
gcc -I/sys/pkg93/include -W -Wall -pedantic -g   -c trurl.c
gcc -I/sys/pkg93/include -W -Wall -pedantic -g   -o trurl trurl.c -Wl,-zrelro -L/sys/pkg93/lib -Wl,-R/sys/pkg93/lib -L/usr/lib -Wl,-R/usr/lib -L/usr/lib -L/sys/pkg93/lib -L/sys/pkg93/lib -L/sys/pkg93/lib -L/sys/pkg93/lib -lcurl
```
Probably OpenBSD's make doesn't have this fallback behaviour.

So, yeah, even though this happens to "work" on standard BSD make, it is really not working as intended.

Nevermind.

nekobbbbbbit:
I didn't think it was a big deal personally. The build system is incredibly simple, ensuring compatibility never hurts and allows me to skip the patch upstream.

Either way, I have created the [patch for now](https://marc.info/?l=openbsd-ports&m=168160930027203&w=2)

Makefile
```diff
@@ -11,6 +12,7 @@ MANDIR ?= $(DESTDIR)$(PREFIX)/share/man/man1
 INSTALL = install
 
 $(TARGET): $(OBJS)
+	$(CC) $(CFLAGS) $(OBJS) -o $(TARGET) $(LDLIBS)
```
emanuele6, line 15: This should not be using `CFLAGS`
```suggestion
	$(CC)  $(LDLIBS) $(OBJS) -o $(TARGET)
```

nekobbbbbbit:
> This currently breaks the sanitizer CI job.

fixed

bagder:
Thanks!

nekobbbbbbit:
When you get a chance/finish up requests, create a new tag so I can update the OBSD port. I also need to implement the python3 test

PR #145, opened 2023-04-16
Makefile: switch to using test.py

bagder:


bagder:
@havbon it seems this does not work very well on cygwin?

havbon:
Thats really strange, it worked for me. I am not available atm so i cant test.

Could you try to remove "3" from Python in the makefile? "@python test.py"

bagder:
> Could you try to remove "3" from Python in the makefile? "@python test.py"

Not easily. On debian:
~~~
$ python
bash: python: command not found

havbon:
Hmm okay, ill do some testing in a bit. I was thinking maybe adding python3 as a package on cygwin?

havbon:
adding `python39` to cygwin packages in `makefile.yml` worked.

`packages: curl, libcurl-devel, libcurl4, make, gcc-core, python39`

PR #147, opened 2023-04-16
--iterate

### LINKED ISSUE #40: Loop over multiple --sets

bagder:
With docs and tests

Alternative take to #117 
Fixes #40


jacobmealey:
this seems cleaner than the PR I opened! one little small note, you never free `o.iter_list` so it's left dangling.

bagder:
... which reminds me we need to get valgrind running the tests!

PR #149, opened 2023-04-17
offer {puny:host} and {puny:url} for --get

### LINKED ISSUE #64: Getting the punycode version
Due to to [how crazy IDN is](https://daniel.haxx.se/blog/2022/12/14/idn-is-crazy/) the only way to be able to get a canonical version of a URL with IDN is to convert it to its punycode version, which I figure is a valid reason for trurl to offer this feature: extract the host as punycoded or show the entire URL with the host name punycoded.

But how to expose this power to the CLI ?

My initial thinking is to use another symbol for the `--get` variables. We already do `{:var}` for *not* URL decoding, and we could use something else than colon to ask for the punycoded version. Maybe '*' ?  Then you can ask for both the plain name and the punycoded one in the same command line: `--get '{host} {*host}'`

Thoughts? Better ideas?

bagder:
Added two tests. Added docs.

Fixes #64

### DIFF
--- a/URL-QUIRKS.md
+++ b/URL-QUIRKS.md
@@ -42,6 +42,12 @@ Implemented in libcurl 7.81.0. Before this, the source formatting was kept.
 
 Run-time requirement.
 
+## `CURLU_PUNYCODE`
+
+Added in libcurl 7.88.0.
+
+Build-time requirement.
+
 ## Accepting % in host names
 
 The host name parser has been made stricter over time, with the most recent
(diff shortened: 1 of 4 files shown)


bagder:
Unfortunately no CI runs a new enough libcurl version to actually test this feature now, they just skip the tests...

bagder:
It could also be noted that this method does not offer the punycode version in the JSON output. Which is a bit weak.

PR #152, opened 2023-04-17
Makefile: fix build with GNU Make 3.x

ZhongRuoyu:
The `!=` operator is not supported by GNU Make 3.x, shipped by macOS.

An implication of this is that the `curl-config` result will not be stored. Given that there is currently only one `.c` source file, it shouldn't make much difference.

```console
$ bmake
cc  $(curl-config --cflags) -W -Wall -pedantic -g -c trurl.c -o trurl.o
cc trurl.o -o trurl $(curl-config --libs)
$ bmake clean
rm -f trurl.o trurl
$ gmake
cc $(curl-config --cflags) -W -Wall -pedantic -g   -c -o trurl.o trurl.c
cc trurl.o -o trurl $(curl-config --libs)
$ gmake clean
rm -f trurl.o trurl
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,7 @@
 TARGET = trurl
 OBJS = trurl.o
-LDLIBS != curl-config --libs
-CFLAGS != curl-config --cflags
-CFLAGS += -W -Wall -pedantic -g
+LDLIBS = $$(curl-config --libs)
+CFLAGS = $$(curl-config --cflags) -W -Wall -pedantic -g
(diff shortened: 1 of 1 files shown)


bagder:
@nekobbbbbbit thoughts?

bagder:
Thanks!

nekobbbbbbit:
> @nekobbbbbbit thoughts?

Looks good! Didn't know about the GNU Make version difference... now everyone is happy :)

PR #157, opened 2023-04-18
added --with-valgrind to test.py

### LINKED ISSUE #148: optionally run test.py with valgrind
We should provide a method to run the tests (in the CI) on Linux with valgrind.

A crude patch that makes it:
~~~diff
diff --git a/test.py b/test.py
index 05473b2..65d21ab 100644
--- a/test.py
+++ b/test.py
@@ -9,11 +9,11 @@ from dataclasses import dataclass, asdict
 
 TESTFILE = "./tests.json"
 if sys.platform == "win32" or sys.platform == "cygwin":
     BASECMD = "./trurl.exe"  # windows
 else:
-    BASECMD = "./trurl"  # linux
+    BASECMD = "valgrind -q ./trurl"  # linux
 
 RED = "\033[91m"  # used to mark unsuccessful tests
 NOCOLOR = "\033[0m"
 TAB = "\x09"

jacobmealey:
this also makes multiple arguments passable to `test.py` 
Also added make test-memory to makefile and added it to the ci stuffs

fixes #148

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -14,8 +14,11 @@ jobs:
       matrix:
         build:
         - name: default
+          install_packages: valgrind
+          test: test-memory
         - name: clang sanitizers
           install_packages: clang
+          test: test
           make_opts: >
             CC=clang
             CFLAGS="-fsanitize=address,undefined,signed-integer-overflow -Wformat -Werror=format-security -Werror=array-bounds -g"
@@ -39,7 +42,7 @@ jobs:
       run: ./trurl -v
 
     - name: test
-      run: make test
+      run: make ${{matrix.build.test}}
 
   build-cygwin:
     runs-on: windows-latest
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,9 @@ clean:
 test:
 	@$(PYTHON3) test.py
 
+test-memory:
+	@$(PYTHON3) test.py --with-valgrind
+
 .PHONY: checksrc
 checksrc:
 	./checksrc.pl trurl.c version.h
--- a/test.py
+++ b/test.py
@@ -9,6 +9,8 @@ from dataclasses import dataclass, asdict
 
 TESTFILE = "./tests.json"
 BASECMD = "./trurl"
+VALGRINDTEST = "valgrind"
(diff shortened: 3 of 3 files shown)


jacobmealey:
I think the Clang Sanitizers Job is failing because you aren't supposed to run Valgrind with `fsanitize=address`. I will try and fix it tomorrow

bagder:
> I think the Clang Sanitizers Job is failing because you aren't supposed to run Valgrind with `fsanitize=address`. I will try and fix it tomorrow

That is correct. The sanitizer jobs should not use valgrind.

.github/workflows/makefile.yml
```diff
@@ -40,6 +44,9 @@ jobs:
 
     - name: test
       run: make test
+    
+    - name: test-memory
+      run: make test-memory
```
bagder, line 49: There is no need to run the tests twice. They should run once, with or without valgrind imho.

.github/workflows/makefile.yml
```diff
@@ -43,10 +41,14 @@ jobs:
...
-    
-    - name: test-memory
-      run: make test-memory
+      run: |
+        if [ "${{matrix.build.name}}" = "clang sanitizers" ]
+        then 
+          make test
+        else 
+          make test-memory
+        fi
```
bagder, line 50: How about just make it `make ${{matrix.build.test}}` and set the `test` variable above in the matrix? It is not very nice to depend on the name of a particular build.

jacobmealey: :smile:  now that's clever! Will do.

.github/workflows/makefile.yml
```diff
@@ -28,7 +30,8 @@ jobs:
       run: |
         sudo apt-get update
         sudo apt-get install libcurl4-openssl-dev ${{ matrix.build.install_packages }}
-
+    - name: install valgrind
+      run: sudo apt-get install valgrind
```
bagder, line 34: move the valgrind install to `install_packages` in the matrix, then only the test-memory build(s) need to install it

bagder:
Unfortunately the python code now merge conflicts with a recent merge we did.

bagder:
Thanks!

PR #158, opened 2023-04-18
Respect CFLAGS, CPPFLAGS in Makefile, fix LDFLAGS order

thesamesam:
The implicit rule handled this correctly but a7f9ca9c42c2cf29cc84becb38b56082898868d4 did away with that.

Add CPPFLAGS (for the C preprocessor) and CFLAGS explicitly, and then shift LDFLAGS earlier to be before the object list, which is needed for e.g. -Wl,--as-needed to work correctly.

Fixes: a7f9ca9c42c2cf29cc84becb38b56082898868d4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ INSTALL ?= install
 PYTHON3 ?= python3
 
 $(TARGET): $(OBJS)
-	$(CC) $(OBJS) -o $(TARGET) $(LDLIBS) $(LDFLAGS)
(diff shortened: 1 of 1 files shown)


thesamesam:
Actually, there's a few other broken things here. The rule for linking vs compiling is inconsistent and `!=` should've been used as that's the POSIX-standardised mechanism for `$(shell ...`). It also still doesn't respect CFLAGS.

I'm going to leave this for now unless there's interest in a Meson port (which would be rather tiny).

PR #163, opened 2023-04-19
Makefile: make  test  target depend on trurl

emanuele6:
To avoid running tests without compiling first :D

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ clean:
 	rm -f $(OBJS) $(TARGET)
 
 .PHONY: test
-test:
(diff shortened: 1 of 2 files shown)


emanuele6:
Added an additional commit that makes the test runner print failed tests in a cleaner and less ambiguous way

emanuele6:
It probably makes more sense to make `install` not depend on trurl; you always want to run install as root, and you probably never want to compile trurl as root.

bagder (approved):

bagder:
Thanks!

PR #172, opened 2023-04-26
--help exits with 0

jacobmealey:
trurl was exiting with 1 when running `--help`, this doesn't seem common for most programs so this PR has it return 0.


bagder (approved):

bagder:
Thanks!

PR #173, opened 2023-04-26
align the -h output nicer

bagder:
Make the long options align under each other. Gives the list a "calmer" look I think.


krishean:
Looks like you accidentally reintroduced the function returning 1 instead of 0

I just noticed now that curl prints help to stdout, while trurl was printing it to stderr, and this makes it consistent

PR #174, opened 2023-04-26
Makefile: make test-memory depend on $TARGET

enno-t:
Since we need an actual compiled trurl to run the valgrind tests, make test-memory depend on $(TARGET).

This also adds a .PHONY to test-memory, since test, checksrc, clean and install all have those.


bagder (approved):

bagder:
Thanks!

PR #177, opened 2023-04-26
added automatic build and upload of artifacts

fragolinux:
it will automatically create a release when a new tag is pushed

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,9 +6,9 @@ on:
       - "v*"
 
 jobs:
-  build-linux:
+  build-ubuntu:
     runs-on: ubuntu-latest
-    name: "Linux Build"
+    name: "Ubuntu Build"
 
     strategy:
       matrix:
@@ -43,17 +43,17 @@ jobs:
       run: make ${{matrix.build.test}}
 
     - name: rename
-      run: mv trurl trurl-linux
+      run: mv trurl trurl-ubuntu
 
     - name: artifacts
       uses: actions/upload-artifact@v3
       with:
-        name: Linux
-        path: ./trurl-linux
+        name: Ubuntu
+        path: ./trurl-ubuntu
 
   build-cygwin:
-    runs-on: windows-latest
-    name: "Windows Build"
+    runs-on: cygwin-latest
+    name: "Cygwin Build"
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


fragolinux:
probably you'd like to rename "windows" in "cygwin", as it's that, really (i just extended your own github action with the release part)

bagder:
I presume these Windows build binaries only work if you have cygwin installed?

.github/workflows/release.yml
```diff
@@ -0,0 +1,132 @@
+name: Release CI
+
+on:
+  push:
+    tags:
+      - "v*"
```
bagder, line 6: Is this supposed to match the tags for which to make releases? They are named `trurl-*` now...

fragolinux:
> I presume these Windows build binaries only work if you have cygwin installed? Then I agree they should be named cygwin.

They're just the same builds already in original automation, just added the release and publish part...

bagder:
We should probably also call them "ubuntu" rather than just "linux".

fragolinux:
Rename them yourself? Or do you want me to do? Just swap windows for cygwin

bagder:
> > I presume these Windows build binaries only work if you have cygwin installed? Then I agree they should be named cygwin.
> 
> They're just the same builds already in original automation, just added the release and publish part...

I don't think that answers my question though?

.github/workflows/release.yml
fragolinux: This automation is triggered only when you push a git tag, in my case I pushed v0.6.0, it's a standard way in gh actions, but you can modify as you want... It makes sense to me, as this way the binaries are created just on a new release, not always

fragolinux:
Lol, I'm just using what you yourself had made, I just renamed (wrongly) cygwin to windows, before realising that without the dll they don't work, of course

fragolinux:
> We should probably also call them "ubuntu" rather than just "linux".

I didn't even take a look on HOW did you build them: if it's a static build, with no dependencies, this should work everywhere on any linux, I'm not able of help you on this...

bagder:
> I'm not able of help you on this...

But I talked about if not a rename would make sense. I did not ask you to change the build.

fragolinux:
> > I'm not able of help you on this...
> 
> But I talked about if not a rename would make sense. I did not ask you to change the build.

It makes sense if these are, really, just ubuntu builds... If the biraries work on other distros, it's good to have them called linux, otherwise better rename them to ubuntu (really, a specific one, too...)

For the windows ones, better rename windows to cygwin

But I see a new pr of someone who probably made native windows builds, too

fragolinux:
i renamed both linux-->ubuntu and windows-->cygwin

bagder:
I think the biggest need for this is for native Windows builds, and I also did not manage to get this to work in a local test repo of mine. I think we can put this on hold for now.

fragolinux:
as windows was never a priority for me, and was in your own code (if working or not, you know...), why not just incorporate the PR parts relative to the *nix builds, and strip out the win part? and my builds were perfectly fine as you can see [here](https://github.com/fragolinux/trurl/actions/runs/4817324474), just the win part failed, as expected (i don't know how to fix them nor interested into)

PR #178, opened 2023-04-26
added support for {raw:port} and raw_port in json output

### LINKED ISSUE #170: ability to get told the port number was not provided?
Right now trurl provides the default port number if none was provided and trurl knows the URL scheme. Like for `https://example.com` it shows `443`.

I can imagine that there might be use cases where you actually rather want to know that no port number was present in the URL. If so, how should we ask for that information?

jacobmealey:
This PR adds support for differentiating between default ports and explicitly defined ports.
Closes #170

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -85,6 +85,7 @@ static const struct var variables[] = {
   {"options",  CURLUPART_OPTIONS},
   {"host",     CURLUPART_HOST},
   {"port",     CURLUPART_PORT},
+  {"raw_port", CURLUPART_PORT},
   {"path",     CURLUPART_PATH},
   {"query",    CURLUPART_QUERY},
   {"fragment", CURLUPART_FRAGMENT},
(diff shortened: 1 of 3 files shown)


trurl.c
```diff
@@ -85,6 +85,7 @@ static const struct var variables[] = {
   {"options",  CURLUPART_OPTIONS},
   {"host",     CURLUPART_HOST},
   {"port",     CURLUPART_PORT},
+  {"raw_port", CURLUPART_PORT},
```
bagder, line 88: By adding this entry here, you also now make it possible to set and get `raw_port` which I'm not sure is a good idea... maybe the JSON output code can just get some special treatment?

jacobmealey: ahh okay! I didn't think of that, thanks for pointing it out! I should be able to push up something cleaner by the end of the day.

jacobmealey:
Yes, I most certainly can do that!

bagder (approved):

bagder:
Thanks!

PR #186, opened 2023-05-03
declaration of 'i' shadow a previous local

eh-san:
declaration of 'i' shadow a previous local (in lines 702 & 735)
at this moment can't cause any problems but in the future, they may!


bagder (approved):

bagder:
Thanks!

PR #188, opened 2023-05-04
Use feature list for tests

### LINKED ISSUE #184: Tests fail with out IDN
It is possible to have a recent version of curl installed without IDN support, so the tests fail.

I was playing around with getting trurl to run on a fresh FreeBSD install, and the curl version packaged by Freebsd didn't ship with  IDN enabled. Note that it is curl 8.0.1. #164 added support for skipping tests based on the minimum version, but IDN (and other features) may need to be explicitly defined for the punycode tests.

jacobmealey:
Fixes #184 
I used the predefines to determine the names of the features, but I feel like it doesn't provide very much insight into what features are added so if folks have a better naming idea then that would be swell. I also only modified the punycode tests because those were the only ones failing, but it may be useful/cleaner to have all of them use the feature names instead of the version number.

### DIFF
--- a/test.py
+++ b/test.py
@@ -8,7 +8,6 @@ import shlex
 from subprocess import PIPE, run
 from dataclasses import dataclass, asdict
 from typing import Any, Optional
-from packaging.version import parse as parse_version
 
 PROGNAME = "trurl"
 TESTFILE = "tests.json"
@@ -124,11 +123,7 @@ def main(argc, argv):
             stdout=PIPE, stderr=PIPE,
             encoding="utf-8"
         )
-        version_re = r' libcurl\/([\w\.\-]+) \[built\-with ([\w\.\-]+)\]$'
-        match = re.findall(version_re, output.stdout)
-        libcurl_runtime, libcurl_buildtime = match[0]
-        libcurl_runtime = parse_version(libcurl_runtime)
-        libcurl_buildtime = parse_version(libcurl_buildtime)
+        features = output.stdout.split('\n')[1].split()[1:]
 
         with open(path.join(baseDir, TESTFILE), "r") as file:
             allTests = json.load(file)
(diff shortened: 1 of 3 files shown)


tests.json
```diff
@@ -1615,7 +1615,7 @@
                 "{puny:url}"
             ]
         },
-        "minbuildtime": "7.88.0",
+        "requiredfeatures": ["punycode"],
```
bagder, line 1618: Can we maybe use something shorter? What about just "required" ?

bagder (approved):
I like this, but I would like to see it taken one step further and remove all the version number checks and convert even the run-time stuff into named features that the tests can depend upon!

jacobmealey:
Okay, excellent I will get on that now!

test.py
```diff
@@ -124,11 +124,7 @@ def main(argc, argv):
             stdout=PIPE, stderr=PIPE,
             encoding="utf-8"
         )
-        version_re = r' libcurl\/([\w\.\-]+) \[built\-with ([\w\.\-]+)\]$'
-        match = re.findall(version_re, output.stdout)
-        libcurl_runtime, libcurl_buildtime = match[0]
-        libcurl_runtime = parse_version(libcurl_runtime)
-        libcurl_buildtime = parse_version(libcurl_buildtime)
```
emanuele6, line 131: `packaging.version` is not a standard python module (I had to install it manually to be able to run tests).

Since this dependency is no longer used, can you please remove the import above?
```python
from packaging.version import parse as parse_version
```

trurl.c
```diff
@@ -184,6 +187,32 @@ static void show_version(void)
   curl_version_info_data *data = curl_version_info(CURLVERSION_NOW);
   fprintf(stdout, "%s version %s libcurl/%s [built-with %s]\n",
           PROGNAME, TRURL_VERSION_TXT, data->version, LIBCURL_VERSION);
+  /* puny code isn't guaranteed based on the version, so it must be polled
+   * from libcurl */
+  bool supports_puny = false;
+#ifdef SUPPORTS_PUNYCODE
+  const char *const *feature_name = data->feature_names;
+  while(*feature_name && !supports_puny) {
+    supports_puny = !strncmp(*feature_name, "IDN", 3);
```
emanuele6, line 196:
```suggestion
    supports_puny = !strcmp("IDN", *feature_name);
```
This seems wrong, it will make `trurl` think libcurl has the "IDN" feature even if `*feature_name` is actually "IDNOSAUR" or some other string that starts with "IDN".
Why are you using this?

jacobmealey: Emanuele, I strive to have your ability to see edge cases in code. I had no reason to do it one way or another, I was just thinking "IDN is only 3 characters so why would I check more?" Thank you :)

emanuele6: Thank you :)

> "IDN is only 3 characters so why would I check more?"

Note that `strncmp("IDN", *feature_name, 4)` would have been equivalent to `strcmp("IDN", *feature_name, 3)`, but 3 isn't, it is effectively checking if `*feature_name` starts with "IDN".
Also note that there is no benefit in using `strncmp("foo", str, 4)` over `strcmp("foo", str)`, `strcmp` stops at the first NUL (`'\0'`) character, and `"IDN"[4] == '\0'`.

jacobmealey: Ahh, that makes sense. my embedded systems professor told us to use strncmp whenever possible because it may go off the end of arrays (if there is no null terminator), so it's mostly force of habit and I haven't done lots of string manipulation/parsing stuff like what we're doing. but hey, now I know lol

emanuele6:
I think is ok to merge now, maybe squash your commits

bagder:
Thanks!

PR #189, opened 2023-05-07
print python error output to stderr

eh-san:
Prints Python standard output in `stdout` and error output in `stderr` so we can separate stdout and stderr from each other and get only one of them like this:  `make test 2> error_list.txt `


jacobmealey (commented):
This looks good to me. The only thing I could think to add is passing verbose printing as a flag.

eh-san:
Thanks!
Ok I'm going to close this PR and open another one that can be rebased and have verbose flag @jacobmealey

Issue #194, opened 2023-05-17
Add more options for configuring curl_url_set()


jacobmealey:
I was thinking it may be useful to have `--verify` be able to accept parameters for what qualifies as a valid URL. I was thinking something like being able to say if no scheme is set then it's not a valid URL. Obviously, other parameters may be chosen as well. 

Hopefully below will help clarify what I'm talking about. 
```
$ trurl "sample" --verify 
http://sample
$ trurl "sample" --verify "scheme"
trurl error: No scheme set
```

emanuele6:
I don't understand what `--verify scheme` is supposed to do.

If `sample` was meant to be the host name, you should have used `-s 'host=sample'` or `-s "host=$sample"`, and you would have gotten:
```
$ trurl -s 'host=sample'
trurl error: not enough input for a URL
trurl error: Try trurl -h for help
```
With `trurl sample` you are specifying `sample` as a URL, and libcurl will parse it as a URL.

Can you specify an actual use case for this to clarify what you want it to do? Why did you think this would be useful?

---

In any case, in my opinion, we should definitely not implement wacky optional option arguments like those. If you want optional arguments, use something like `--verify=foo` to specify the optional argument or just use another option.

A `--verify` that may or may not interpret the next argument as an option argument, or that must be specified at the end if you don't want to pass the optional argument (making it impossible to use with `--`) are just bad non-script-friendly option styles.

jacobmealey:
Sorry for the confusion, `scheme` here would be the preceding `http://`, and `sample` is a generic URL. The use case I'm thinking of is a script for extracting all the "valid" URLs in a large text file. trurl has a lot of the mechanics in place for something like this, but it would be kind of roundabout, so it would be nice to have something that can verify the string is a URL that meets some requirements (like having a preceding HTTP or something). It might make more sense to do it with something similar to `--get {raw:port}`. Let me know if you have any more questions, I feel like I am missing some keywords in my explanation.

emanuele6:
> Sorry for the confusion, `scheme` here would be the preceding `http://`, and sample is a generic URL.

Can you provide an example command line of what you mean with`http://` as scheme?
You mean like `trurl -u "$url" --verify=http://`?

> The use case I'm thinking of is a script for extracting all the "valid" URLs in a large text file.

This doesn't really help much understand what you mean.

`trurl` already discards unparsable URLs in input files:
```bash
$ cat foo.txt
https://example.org
file:///home/emanuele6/./foo/..
\\
foo
[...]

PR #195, opened 2023-05-18
trurl: add a new --no-guess-scheme option to disable scheme guessing

emanuele6:
Ref: #194


emanuele6:
The memory leak is detected by the [CI job](https://github.com/curl/trurl/actions/runs/5009209706/jobs/8977903588?pr=195) is unrelated.
Fixed by #196

bagder (approved):

PR #196, opened 2023-05-18
trurl: fix URL handle leak for invalid URLs

emanuele6:
Before this patch, when trurl encountered a invalid URL, it skipped it without freeing the URL handle.
This was not discovered earlier because all the tests with invalid URLs were using `--verify` so trurl aborted immediately when encountering invalid URLs, and valgrind couldn't detect the memory leak.


bagder (approved):

bagder:
Excellent!

PR #197, opened 2023-05-18
--url-file: skip long and empty lines instead of successfully aborting

emanuele6:
This fixes a regression from aa62bb1 that caused trurl to abort successfully at the first empty (but not blank) line instead of simply skipping it.

This patch also makes trurl skip too long lines instead of successfully aborting when encountering one.

It also avoids calling fgets() after EOF, and reports read errors caused by getc()/fgets() if any.

I added a testfiles/ directory in which to add --url-file files to use in tests, and some tests that check that empty lines are ignored, long lines are reported and skipped, and spaces and tabs, and final carriage return are trimmed off the end lines.


bagder (approved):

bagder:
tests.json conflicts now though, do you mind fixing that?

emanuele6:
Done

tests.json
```diff
@@ -1729,6 +1729,32 @@
             "stdout": "emanuele6://curl.se/trurl\nhttps://example.org/\n",
             "returncode": 0,
             "stderr": "trurl note: No host part in the URL []\n"
+    },
```
bagder, line 1732: There's a missing close brace here isn't it?

bagder (approved):

emanuele6:
The commit had the wrong name for some reason :/

PR #198, opened 2023-05-18
--redirect: don't silently ignore failed redirections.

emanuele6:
If a redirection fails, report the failure to stderr, and go to the next URL, or abort if running with --verify. Instead of silently ignoring the failure and carrying on.

This patch also renames urlfromstring() to seturl(), and implements --redirect with it so that curl_url_set() is called with the same flags used to parse base URLs.

Which means the following command will now work:

  $ trurl --accept-space --redirect 'x://foo/a b' curl.se
  x://foo/a%20b

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -1761,5 +1761,35 @@
             "returncode": 0,
             "stderr": "trurl note: Bad scheme [foo]\ntrurl note: Bad scheme [hi]\ntrurl note: Bad scheme [hey]\n"
         }
+    },
+    {
+        "input": {
+            "arguments": [
+                "--accept-space",
+                "--redirect",
+                "https://example.org/a b",
+                "https://curl.se"
+            ]
+        },
+        "expected": {
+            "stdout": "https://example.org/a%20b\n",
+            "returncode": 0,
+            "stderr": ""
+        }
+    },
+    {
+        "input": {
+            "arguments": [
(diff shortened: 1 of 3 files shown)


emanuele6:
Oops, I forgot to remove the test with a space in the URL that I didn't want to push to begin with. =)

It should be alright now... except that the libcurl on the CI, and libcurl 8.1.0 on PC output different error messages... :/

emanuele6:
Now tests can contain `true` or `false` expectectations, `true` expectations are verified by non-empty output, or zero exit codes, `false` expectactions are verified by empty output, or non-zero exit codes.

tests.json
```diff
@@ -1730,5 +1730,35 @@
...
         }
+    },
+    {
+        "input": {
+            "arguments": {
+                "--verify",
+                "--redirect",
+                "https://example.org/a b",
+                "https://curl.se"
+            ]
```
bagder, line 1741:
```suggestion
            }
```

bagder (commented):
Also seems to merge conflict now

emanuele6:
Should be fine now

emanuele6:
Hmm, let me investigate that memory leak...

emanuele6:
oops :S

bagder (approved):

PR #202, opened 2023-05-20
Remove set url

jacobmealey:
As @emanuele6 mentioned in #201 using url with --set should be removed. This PR has trurl report an error and provides the following message: `--set does not support url, use --redirect instead`

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -311,6 +311,8 @@ static void setadd(struct option *o,
 {
   struct curl_slist *n;
   n = curl_slist_append(o->set_list, set);
+  if(!strncmp("url=", set, 4))
+      errorf(ERROR_SET, "--set does not support url, use --redirect instead");
   if(n)
     o->set_list = n;
 }
--- a/trurl.c
+++ b/trurl.c
@@ -312,7 +312,7 @@ static void setadd(struct option *o,
   struct curl_slist *n;
   n = curl_slist_append(o->set_list, set);
   if(!strncmp("url=", set, 4))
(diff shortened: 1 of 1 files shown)


emanuele6:
I think this patch makes `--set 'url=foo` error when options are parsed, but still allows `--set 'url:=foo'` or `--iterate 'url=...'`.

jacobmealey:
Oh, good catch, I didn't even know `url:=...` was possible. I just pushed another commit. I'm not super happy with it, so if you have suggestions let me know. 

Also, congrats on becoming a _collaborator_!

trurl.c
```diff
@@ -320,6 +322,8 @@ static void iteradd(struct option *o,
 {
   struct curl_slist *n;
   n = curl_slist_append(o->iter_list, iter);
+  if(!strncmp("url", iter, 3))
+    errorf(ERROR_SET, "--iterate does not support url");
```
bagder, line 326: ERROR_ITER ?

trurl.c
```diff
@@ -311,6 +311,8 @@ static void setadd(struct option *o,
 {
   struct curl_slist *n;
   n = curl_slist_append(o->set_list, set);
+  if(!strncmp("url", set, 3))
+    errorf(ERROR_SET, "--set does not support url, use --redirect instead");
```
bagder, line 315: Shouldn't this check be done *before* curl_slist_append() is called?

bagder (approved):

bagder:
Thanks!

PR #204, opened 2023-05-20
trurl: restructure JSON, add --default-port, --keep-port, --punycode

emanuele6:
This patch restructures the JSON output with the following format:
```
  {
    "url": // same value as `{url}`
    "parts": {
      ..., // urldecoded URL parts except the URL
    },
    "params": [
       ...
    ],
    // other stuff
  }
```

I decided to remove raw_port for now, since now I think it is less useful than it used to be.

This patch also reworks default output and -g.
`{raw:}` has been removed in favour of `{default:}`.
`{default:}` and `{puny:}` are "get modifiers" like `{:component}`, and may be specified in any order front of a component to modify the curl_url_get() flags used by trurl to get that URL.

`{default:}` makes trurl use CURLU_DEFAULT_PORT, use the libcurl-supported scheme's default port if one is not specified, and `{puny:}` makes trurl use CURLU_PUNYCODE  e.g.

* `{default:port}` will get 80 for `http://example.org/`
* `{puny:host}` will get xn--p-8fa.example.org from `imap://pè.example.org`

`{default:}`, `{puny:}`, and `{:part}` may be specified at the same time in which case multiple additional flags are passed to curl_url_get().

  $ trurl -g '{puny:default:url}' imap://pè.example.org
  imap://xn--p-8fa.example.org:143/

If `--default-port` is used, `{default:}` is implied for every `{}` in the `--get;` and, if `--puny` is implied.

Default output now behaves exactly like  -g '{url}'. (note: previously  -g '{url}'  behaved like the new `{default:url}`, it
       always included, )

This allows more fine tuning of `-g` the output, and default URL output since it was previously not possible to:

* get the URL with explicit default port in the `--json` output
* get the URL without default port in a `--get` format
* get the punycode encoded URL without explicit default port
* get the punycode encoded URL with or without explict port in `--json`

By default, trurl removes redundant explict ports for libcurl-supported schemes, you can use `--keep-port` to inhibit that behaviour.

I fixed a bug that made  -g `{raw:port}`  (or  -g `{port}`) never able to print the default port for libcurl-supported scheme even if the port was explictly specified. (this bug was unique to `{raw:port}`, raw_port in the JSON worked fine.)

I also fixed some inconsistencies in the code, e.g. get() had a output stream variable that was not being propagated to showqkey() that was always outputting to stdout, and failed gets were printing warnings using  printf(stderr, PROGNAME ...)  instead of warnf().

This is only an initial draft, so I have not updated test expectations and the manual yet.

Ref: #201

### DIFF
--- a/trurl.1
+++ b/trurl.1
@@ -50,6 +50,16 @@ encode such occurrences accordingly.
 
 According to RFC 3986, a space cannot legally be part of a URL. This option
 provides a best-effort to convert the provided string into a valid URL.
+.IP "--default-port"
+When set, trurl will use the scheme's default port number for URLs with a known
+scheme, and without an explicit port number.
+
+Note that trurl only knows default port numbers for URL schemes that are
+supported by libcurl.
+
+Since, by default, trurl removes default port numbers from URLs with a known
+scheme, this option is pretty much ignored unless one of \fI--get\fP,
+\fI--json\fP, and \fI--keep-port\fP is not also specified.
 .IP "-f, --url-file [file name]"
 Read URLs to work on from the given file. Use the file name "-" (a single
 minus) to tell trurl to read the URLs from stdin.
@@ -71,8 +81,26 @@ command line.
 The following component names are available (case sensitive): url, scheme,
 user, password, options, host, port, path, query, fragment and zoneid.
 
+\fB{component}\fP will expand to nothing if the given component does
+not have a value.
+
 Components are shown URL decoded by default. If you instead write the
 component prefixed with a colon like "{:path}", it gets output URL encoded.
+
+You may also prefix components with \fBdefault:\fP and/or \fBpuny:\fP,
+in any order.
+
+If \fBdefault:\fP is specified, like "{default:url}" or
+"{default:port}", and the port is not explicitly specified in the URL,
+the scheme's default port will be output if it is known.
+
+If \fBpuny:\fP is specified, like "{puny:url}" or "{puny:host}", the
+"punycoded" version of the host name will be used in the ouptut.
+
+If \fI--default-port\fP is specified, all formats are expanded as if
+they used \fIdefault:\fP; and if \fI--punycode\fP is used, all formats
+are expanded as if they used \fIpuny:\fP. Also note that "{url}" is
+affected by the \fI--keep-port\fP option.
 
 Hosts provided as IPv6 numerical addresses will be provided within square
 brackets. Like "[fe80::20c:29ff:fe9c:409b]".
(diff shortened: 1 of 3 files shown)


emanuele6:
Just need to update the man page now

bagder (approved):

bagder (commented):
This looks good. I just miss the corresponding update of the man page!

emanuele6:
I updated the man page, rebased on master, and reverted f44a586 since it is not necessary anymore

emanuele6:
I updated the man page, rebased on master, and reverted f44a586 since it is not necessary anymore

trurl.1
```diff
@@ -71,9 +81,27 @@ command line.
 The following component names are available (case sensitive): url, scheme,
 user, password, options, host, port, path, query, fragment and zoneid.
 
+\fB{component}\fP will expand to nothing if the given componetnt does
```
bagder, line 84:
```suggestion
\fB{component}\fP will expand to nothing if the given component does
```

bagder (approved):
Looks awesome. Sorry for being a bit slow. I just found a tiny typo.

bagder:
I'm guessing the test now needs a new `required` ?

emanuele6:
I will remove these two tests:
```json
    {
        "input": {
            "arguments": [
                "--accept-space",
                "--redirect",
                "https://example.org/a b",
                "https://curl.se"
            ]
        },
        "expected": {
            "stdout": "https://example.org/a%20b\n",
            "returncode": 0,
            "stderr": ""
        }
    },
    {
        "input": {
            "arguments": [
                "--verify",
                "--redirect",
                "https://example.org/a b",
                "https://curl.se"
            ]
        },
        "expected": {
            "stdout": "",
            "returncode": 9,
            "stderr": true
        }
    },
```

As previously mentioned, now `trurl` always parses URLs with `CURLU_URLENCODE`, and, before 8.1.0, the URL parser was not strict enough and always allowed spaces with `CURLU_URLENCODE`, even when `CURLU_ACCEPT_SPACE` was not used.

> NOTE: Since now trurl always uses CURLU_URLENCODE when parsing URLs,
> tests that reject spaces now require libcurl 8.1.0, because before
> curl/curl@4cfa5bcc (curl 8.1.0), URLs with spaces are not reject spaces
> even if CURLU_ALLOW_SPACES is not provided.

I avoided having to delete these tests earlier by using `"minruntime": "8.1.0",`, but now I can't do that any more.

bagder (approved):

bagder:
:+1:

PR #206, opened 2023-05-25
use stderr in test.py

eh-san:
This is from Ref: https://github.com/curl/trurl/pull/189
separate stdout and stderr from each other, 
so we can get the error only if we want to find out where exactly the error occurred like this: ` make test 1> /dev/null `
Screenshot from 2023-05-25 23-25-49
and also add a `verbose` flag to print verbose if we want to see all the detail: ` python3 test.py --verbose ` 
Screenshot from 2023-05-25 23-22-27

### DIFF
--- a/test.py
+++ b/test.py
@@ -29,7 +29,7 @@ import json
 import shlex
 from subprocess import PIPE, run
 from dataclasses import dataclass, asdict
-from typing import Any, Optional
+from typing import Any, Optional, TextIO
 
 PROGNAME = "trurl"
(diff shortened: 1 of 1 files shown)


bagder:
Thanks @eh-san for this. Can you please rebase, fix the conflicts and force-push again?

eh-san:
Sure, I fixed it
Thanks for reviewing @bagder

bagder:
Thanks!

PR #207, opened 2023-05-26
Use DESTDIR only during install target

rbgarga:
At least on FreeBSD ports framework, DESTDIR is only defined during install stages, what makes it impossible to set proper make environment variables.


bagder (approved):

bagder:
Thanks!

PR #214, opened 2023-06-01
format null as \u0000 when calling with --json 

### LINKED ISSUE #205: Query params values in JSON output have NUL replaced by .
NUL is replaced by `.` also in the JSON output, not just in the output of `-g '{query:}'` and `-g '{query-all:}'`.

Since `JSON` supports NUL in strings, and values in params are URL decoded, I think it should not be replaced.
```diff
 [
   {
     "url": "http://example.org/?q=mr%00smith",
     "parts": {
       "scheme": "http",
       "host": "example.org",
       "path": "/",
       "query": "q=mr%00smith"
     },
     "params": [
       {
         "key": "q",
-        "value": "mr.smith"
+        "value": "mr\u0000smith"
       }
     ]
   }
 ]
```

jacobmealey:
- has `--json` encode a null byte as `\u0000` instead of `.` 
- adds `struct string` and makes `qpairs` and `qpairsdec` use this instead of a char*
- fixes #205

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -902,9 +902,9 @@ struct string *memdupzero(char *source, size_t len)
 {
   struct string *ret = malloc(sizeof(struct string));
   if(!ret)
-      return NULL;
+    return NULL;
   ret->str = malloc(len + 1);
-  if(!ret->str) 
+  if(!ret->str)
     return NULL;
   memcpy(ret->str, source, len);
   ret->str[len] = 0;
@@ -1006,7 +1006,6 @@ static char *addqpair(char *pair, size_t len, bool json)
     free(pdec);
   if(p)
     free(p);
-    
   return ret;
 }
 
--- a/trurl.c
+++ b/trurl.c
@@ -908,8 +908,10 @@ struct string *memdupzero(char *source, size_t len)
   if(!ret)
     return NULL;
   ret->str = malloc(len + 1);
-  if(!ret->str)
+  if(!ret->str) {
+    free(ret);
     return NULL;
+  }
   memcpy(ret->str, source, len);
   ret->str[len] = 0;
   ret->len = len;
@@ -954,6 +956,9 @@ struct string *memdupdec(char *source, size_t len, bool json)
   curl_free(right.str);
   curl_free(left.str);
   struct string *ret = malloc(sizeof(struct string));
+  if(!ret) {
+      return NULL;
+  }
   ret->str = str;
   if(right.str)
     ret->len = right.len;
@@ -996,7 +1001,8 @@ static char *addqpair(char *pair, size_t len, bool json)
   else
     warnf("too many query pairs");
 
-  ret = p->str;
+  if(p)
+    ret = p->str;
   if(pdec)
     free(pdec);
   if(p)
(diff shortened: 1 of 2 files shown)


trurl.c
```diff
@@ -749,7 +755,7 @@ static void jsonString(FILE *stream, const char *in, size_t len,
   for(; i < (unsigned char *)in_end; i++) {
     switch(*i) {
     case '\\':
-      fputs("\\\\", stream);
+      fputs("\\", stream);
```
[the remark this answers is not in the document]
emanuele6: It definitely should not be changed, you are breaking how `\` is outputted. :/

If instead of replacing NUL characters with `\u0000`, you simply leave them unchanged it will work fine.

trurl.c
```diff
@@ -817,18 +823,19 @@ static void json(struct option *o, CURLU *uh)
...
         continue;
       if(!first)
         fputs(",\n", stdout);
       first = false;
       fputs("      {\n        \"key\": ", stdout);
-      jsonString(stdout, qpairsdec[j],
-                 sep ? (size_t)(sep - qpairsdec[j]) : strlen(qpairsdec[j]),
+      jsonString(stdout, qpairsdec[j].str,
+                 sep ? (size_t)(sep - qpairsdec[j].str) :
+                       strlen(qpairsdec[j].str),
```
emanuele6, line 838: `.len`

trurl.c
```diff
@@ -880,92 +889,125 @@ static void trim(struct option *o)
...
+  right.len = 0;
+
+  left.str = strurldecode(source, sep ? (size_t)(sep - source) : len,
+                      (int *)&left.len);
   if(sep) {
     char *p;
     int plen;
-    right = strurldecode(sep + 1, len - (sep - source) - 1, &rightlen);
+    right.str = strurldecode(sep + 1, len - (sep - source) - 1,
+            (int *)&right.len);
```
emanuele6, line 939: Using `(int *)&left.len` doesn't seem great since you are casting a pointer to a larger unsigned int, to a pointer to a smaller signed int. I would rather use a temporary int variable, and then `right.len = len`.

But really, if you only convert decoded values to `struct string`, instead of both decoded values and encoded values, it would be better in my opinion.

Using `struct string` for encoded values is pointless and doens't help in any way.

trurl.c
```diff
@@ -1016,16 +1059,19 @@ static void qpair2query(CURLU *uh, struct option *o)
 /* sort case insensitively */
 static int cmpfunc(const void *p1, const void *p2)
 {
-  return strcasecmp(*(const char **)p1,
-                    *(const char **)p2);
+  int len = (((struct string *)p1)->len) < (((struct string *)p2)->len)?
+             (((struct string *)p1)->len):(((struct string *)p2)->len);
+
+  return strncasecmp((((struct string *)p1)->str),
+                    (((struct string *)p2)->str), len);
```
emanuele6, line 1066: This won't actually compare `struct string`. `strncasecmp` only compares NUL terminated strings, with a maximum size.
`"foo\0bar"` and `"foo\0zap"` will be the same string for this.

bagder (approved):
I think this looks good. Can you have a look at the little merge conflict and I'll be ready to merge!

jacobmealey:
Ahh there we go! it should be all set to merge now bagder, thanks!

PR #217, opened 2023-06-01
--json/--get: .[].parts and {component} should be urldecoded not encoded; add --urlencode

emanuele6:
I accidentally implemented {:component} backwards in https://github.com/curl/trurl/commit/5bc344a704dad9413b9ff2406c163fb471f8e7be

Oops :)

This also caused URL parts in the json output to be urlencoded instead of urldecoded, since code is shared.

---

One of the reason we split URL parts into a separate object was to allow users to apply transformations on a part of the URL using a jq command that maps the parts object to a trurl command that constructs the new modified URL, e.g. like so:
```sh
trurl --json --url "$url" |
jq -r '
  .[].parts |
  (.path | values) |= ascii_upcase |# uppercase the path if it present
  @sh "trurl \([ to_entries[] | "-s", "\(.key)=\(.value)"])"
' |
sh
```

But unfortunately, since the parts are urldecoded, this can't work.

libcurl, intead of returning a URL encoded part even if CURLU_URLDECODE is used like it does for CURLUPART_URL, when you request
```c
  curl_url_get(uh, CURLUPART_QUERY, &query, CURLU_URLDECODE)
```
returns the URL decoded version of the entire query string which is pretty much useless.

`?a%26b&a&b` becomes `?a&b&a&b` which is completely different, so that jq pipeline will run a command that won't reparse the query correctly. And there is not much it can do about that.

To fix this inconvenience, I suggest to add a --urlencode option that works like --punycode and --default-port, and makes {path} behave like {:path} by default (and consequently causes URL parts in the JSON output to be urlencoded).

Now users should be able to just use `--urlencode --json` instead of just `--json`, and `"-s", "\(.key):=\(.value)]"` instead of `"-s", "\(.key)=\(.value)]"` or `"-s", "\(.key):=\(.value | @uri)]"`, and that pipeline should work fine.

I also added tests to verify that the pipeline actually works.

And I also mentioned in the manual that the host will be punycoded in the json output if --punycode is used, which was previously not documented.

### DIFF
--- a/trurl.1
+++ b/trurl.1
@@ -193,6 +193,9 @@ Providing multiple URLs will make trurl act on all URLs in a serial fashion.
 
 If the URL cannot be parsed for whatever reason, trurl will simply move on to
 the next provided URL - unless \fI--verify\fP is used.
+.IP "--urlencode"
+Outputs URL encoded version of components by default when using \fI--get\fP or
+\fI--json\fP.
 .IP "--trim [component]=[what]"
 Trims data off a component. Currently this can only trim a query component.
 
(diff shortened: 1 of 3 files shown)


bagder (approved):
Sound and reasonable logic I approve of!

trurl.1
```diff
@@ -216,41 +219,42 @@ This key exists in every object. It is the complete URL. Affected by
 .IP "parts"
 This key exists in every object, and contains an object with a key for
 each of the settable URL components. If a component is missing, it means
-it is not present in the URL.
+it is not present in the URL. The parts are URL decoded unless \fI--urlencode\fB
```
emanuele6, line 222:
```suggestion
it is not present in the URL. The parts are URL decoded unless \fI--urlencode\fP
```
oops

emanuele6: amended

emanuele6:
I forgot to make `{query:foo}` and `{query-all:bar}` respect `--urlencode`. I fixed that, and added tests for `{:query:foo}` and `--urlencode -g {query:foo}`

PR #221, opened 2023-06-12
Clean up whitespace, spelling/grammar, unused things, scope, shadowing, checks, file opens

pabs3:
- Remove trailing whitespace
- Remove whitespace from arrays
- Split up a long line
- Fix indentation
- Uppercase the manual page document title
- Fix typos
- Remove an unused import
- Remove unused struct member
- Reduce scope of some variables
- Remove unnecessary pointer check
- Avoid using the name of a function for a variable within that function
- Modernise Perl file open calls
- Only check trurl.c for spelling errors once

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -33,24 +33,24 @@ jobs:
         language: ['cpp', 'python']
 
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@v3
+      - name: Checkout repository
+        uses: actions/checkout@v3
 
-    # Initializes the CodeQL tools for scanning.
-    - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
-      with:
-        languages: ${{ matrix.language }}
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: ${{ matrix.language }}
 
-    - name: install libcurl
-      run: |
-        sudo apt-get update
-        sudo apt-get install libcurl4-openssl-dev
+      - name: install libcurl
+        run: |
+          sudo apt-get update
+          sudo apt-get install libcurl4-openssl-dev
 
-    - name: build
-      run: make
+      - name: build
+        run: make
 
-    - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v2
-      with:
-        category: "/language:${{matrix.language}}"
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2
+        with:
+          category: "/language:${{matrix.language}}"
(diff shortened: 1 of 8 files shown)


trurl.1
```diff
@@ -2,7 +2,7 @@
 .\" man -l trurl.1
 .\" Written by Daniel Stenberg
 .\"
-.TH trurl 1 "April 27, 2023" "trurl" "trurl Manual"
+.TH TRURL 1 "April 27, 2023" "trurl" "trurl Manual"
```
bagder, line 5: I disagree

.github/workflows/codeql.yml
```diff
@@ -30,27 +30,27 @@ jobs:
...
+          sudo apt-get update
+          sudo apt-get install libcurl4-openssl-dev
+
+      - name: build
+        run: make
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2
+        with:
+          category: "/language:${{matrix.language}}"
```
bagder, line 56: Why the reindents of the yaml files? They seem to already be using the indent level GitHub itself uses when it generates such yaml files...

checksrc.pl
```diff
@@ -933,7 +933,7 @@ sub scanfile {
 
     checksrc_endoffile($file);
 
-    close(R);
+    close($R);
```
bagder, line 936: Note that `checksrc.pl` is imported from the curl source tree (https://github.com/curl/curl/blob/master/scripts/checksrc.pl) so it might very well be overwritten again at a later point in time when/if we re-import changes from there.

trurl.c
```diff
@@ -733,8 +732,8 @@ static unsigned int set(CURLU *uh,
   unsigned int mask = 0;
   for(node =  o->set_list; node; node = node->next) {
     const struct var *v;
-    char *set = node->data;
-    v = setone(uh, set);
+    char *name = node->data;
+    v = setone(uh, name);
```
bagder, line 736: I don't think this change helps. It's more than a name.

trurl.1
pabs3: On my system, most manual pages follow this convention.

I've removed this commit from my local copy of the branch.

.github/workflows/codeql.yml
pabs3: In `.github/workflows/makefile.yml`, the indentation for the `steps` sections is inconsistent between the `ubuntu` and `cygwin`/`macos` sections. In addition, `yamllint` by default recommends the extra indentation style, so I elected to make all the files consistent with that.

If you prefer I can reformat the `cygwin`/`macos` sections of `.github/workflows/makefile.yml` to use the GitHub style indentation for `steps`.

checksrc.pl
pabs3: Would you like me to submit a PR for forwarding the change there, or can you copy the commit over?

trurl.c
pabs3: I see the `setone` function calls that parameter `setline`, using that instead.

bagder (approved):

PR #225, opened 2023-06-22
trurl: run  --trim query  before  --append query

emanuele6:
Before this patch there was not a way to add a query parameter to a URL, removing the previous value if the parameter was already present in the base URL.

I tried to use:
```sh
x () {
  trurl --verify --trim query=foo --append "query=foo=$2" -- "$1"
}
```

But that didn't work since  --trim query  runs after --append query, so the new value value added with --append also gets trimmed:

    $ x 'https://example.org/?foo=hello&baz=boo' 'howdy'
    https://example.org/?baz=boo

This patch make trurl run  --trim  before  --append  so that using --trim query=foo --append query=foo=bar  works to set foo to bar overwriting previous value of foo.

    $ x 'https://example.org/?foo=hello&baz=boo' 'howdy'
    https://example.org/?baz=boo&foo=howdy

I added a test for this, and I also added a test that tests that --sort-query  and  --trim query  work together, since I noticed it was missing.


bagder (approved):
Fair!

PR #229, opened 2023-08-07
trurl: accept \* as a trim name to trim a literal asterisk name

### LINKED ISSUE #227: unable to selectively remove query parameters with asterisk in them
Query parameters containing the `*` character cannot be effectively targeted as the wildcard matches other parameters as well. I suggest identifying `%2a` with `*`, which currently would only match a literal `%2a` in the original URI, as a workaround.

bagder:
Added test case. Added docs.

Reported-by: Anselm Schüler
Fixes #227

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -2098,5 +2098,20 @@
                 }
             ]        
         }
+    },
+    {
+        "input": {
+            "arguments": [
+                "--url",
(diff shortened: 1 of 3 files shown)


bagder:
@schuelermine what do you think?

schuelermine:
Cool, thank you! That works for my use-case

bagder:
Excellent, thanks!

PR #231, opened 2023-08-14
Puny2idn

jacobmealey:
This PR adds a few new features around converting punycode to IDN. specifically `--as-idn` and `idn:` option in the get syntax.

The development for this work was done using the [latest version of curl from master.]( https://github.com/curl/curl/commit/67e9e3cb1ea498cb94071dddb7653ab5169734b2) If you link against this version you should be able use these features (though trurl works with out them). 

There is a large uptick in memory related issue - either double frees or curl_free not cleaning everything up. I addressed some of them but I am having trouble identifying the others, so this may take more investigation.

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -2122,7 +2122,7 @@
                 "--as-idn"
             ]
         },
-        "required": ["punycode2unicode"],
+        "required": ["punycode2idn"],
         "expected": {
             "stdout": "http://räksmörgås/\n",
             "stderr": "",
@@ -2138,9 +2138,24 @@
                 "{idn:host}"
             ]
         },
-        "required": ["punycode2unicode"],
+        "required": ["punycode2idn"],
         "expected": {
             "stdout": "räksmörgås\n",
+            "stderr": "",
+            "returncode": 0
+        }
+    },
+    {
+        "input": {
+            "arguments": [
+                "--url",
+                "http://xn--abcdef/",
+                "--as-idn"
+            ]
+        },
+        "required": ["punycode2idn"],
+        "expected": {
+            "stdout": "http://xn--abcdef/\n",
             "stderr": "",
             "returncode": 0
         }
--- a/tests.json
+++ b/tests.json
@@ -2150,13 +2150,29 @@
             "arguments": [
                 "--url",
                 "http://xn-----/",
-                "--as-idn"
+                "--as-idn",
+                "--quiet"
             ]
         },
         "required": ["punycode2idn"],
         "expected": {
             "stdout": "http://xn-----/\n",
             "stderr": "",
+            "returncode": 0
+        }
+    },
+    {
+        "input": {
+            "arguments": [
+                "--url",
+                "http://xn-----/",
+                "--as-idn"
+            ]
+        },
+        "required": ["punycode2idn"],
+        "expected": {
+            "stdout": "http://xn-----/\n",
+            "stderr": "trurl note: Error converting url to IDN [Bad hostname]\n",
             "returncode": 0
         }
     }
--- a/tests.json
+++ b/tests.json
@@ -2155,8 +2155,8 @@
         },
         "required": ["punycode2idn"],
         "expected": {
-            "stdout": "",
-            "stderr": "trurl note: invalid url [Bad hostname]\n",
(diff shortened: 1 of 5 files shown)


tests.json
```diff
@@ -2113,5 +2113,36 @@
...
+                "-g",
+                "{idn:host}"
+            ]
+        },
+        "required": ["punycode2unicode"],
+        "expected": {
+            "stdout": "räksmörgås\n",
+            "stderr": "",
+            "returncode": 0
+        }
```
bagder, line 2146: Maybe also do a test with a host name that starts with `xn--` but isn't a valid punycode. Like perhaps `xn--abcdef` or something.

trurl.1
```diff
@@ -109,15 +109,20 @@ not have a value.
...
 "{default:port}", and the port is not explicitly specified in the URL,
 the scheme's default port will be output if it is known.
 
 If \fBpuny:\fP is specified, like "{puny:url}" or "{puny:host}", the
-"punycoded" version of the host name will be used in the output.
+"punycoded" version of the host name will be used in the output. This
+option is mutually exclusive with \fBidn:\fP.
+
+If \fBidn:\fP is specified like "{idn:url}" or "{idn:host}", the International
+Domain Name version of the host name will be used in the output. This
```
bagder, line 124:
```suggestion
Domain Name version of the host name will be used in the output if it is provided as a correctly encoded punycode version. This
```

trurl.c
```diff
@@ -227,7 +232,10 @@ static void show_version(void)
   fprintf(stdout, "url-strerror ");
 #endif
 #ifdef SUPPORTS_NORM_IPV4
-  fprintf(stdout, "normalize-ipv4");
+  fprintf(stdout, "normalize-ipv4 ");
+#endif
+#ifdef SUPPORTS_PUNY2UNICODE
+  fprintf(stdout, "punycode2unicode");
```
bagder, line 238: Maybe `punycode2idn` as we use the term "idn" for this elsewhere?

tests.json
jacobmealey: I've added a test for this, and it behaves different then I would expect. Not sure if it's something on the trurl side or something on the new libcurl apis. 
```
$ trurl --url "http://xn--abcdef/" --as-idn
http:///
$ trurl --url "http://xn--abcdef/" -g "{idn:host}"
[empty line] 
$
```

bagder: I believe the API will (or at least should) return an error if you try to set a bad host name, which this should count as. Does it not?

bagder: I made https://github.com/curl/curl/pull/11674 to clarify what to expect on bad puny code.

jacobmealey:
@bagder Currently this branch returns `trurl note: invalid url [Bad hostname]` if the curl_url_get returns CURLUE_BAD_HOST. If the user enters `trurl --url "http://xn-----/" --as-idn` should it return `http://xn-----/` or some error message like `trurl note: invalid punycode` ?

bagder:
> I think it makes sense to return `http://xn-----/`.

I agree. We can possibly output a warning (?) but I think it makes sense to treat `--as-idn` as convert-if-possible.

Does it make sense with how the libcurl API works or do we need to poke anything in that end as well?

jacobmealey:
libcurl should be fine as is, I have a commit on my personal computer that behaves this way that I can push up when I get home. Currently libcurl reports an error on conversion which it should, and trurl  (on the commit I haven't pushed up) is handling the error by simply returning original url. I can make sure it throws a warning as well.

bagder:
> I can make sure it throws a warning as well.

I'm not entirely sure on that warnings is a good idea, but I figure we can try it. I suspect that if we introduce "warnings" for some things we probably also need to provide a way to hush them up...

jacobmealey:
hm... we already have `warnf` could we have some function like `trurl_warnf(option *o, char *fmt, ...)` and have an option for supressing warnings?

```c
trurl_warnf(option *o, char *fmt, ...) 
{
  ... 
  if(!o->supress_warnings) {
    warnf(fmt, ap);
  }
}
```

bagder:
Yes, I think that is very sensible.

bagder (approved):

PR #234, opened 2023-08-30
General Cleanup

jacobmealey:
`struct option *` was referred to as both `o` and `op` this patch makes it so all occurrences of it to `o`. I also removed some unneeded newlines in if statements.

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -448,13 +448,11 @@ static int getarg(struct option *o,
     urlfile(o, arg);
     *usedarg = true;
   }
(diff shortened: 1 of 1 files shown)


bagder (approved):
Sorry this has taken me forever. Can you please take a look at the merge conflict and then we should be good to go!

jacobmealey:
No worries, it seemed like a hectic month or two at the Curl project with that CVE. Anyway, this should be all set, Thanks!

PR #235, opened 2023-09-02
Trurl_errorf

jacobmealey:
Added trurl_errorf which handles cleanup of `struct option` before exiting but after printing in case whats being printed must be cleaned up. It is based on #234 so that should be merged before this. Note that this (should) fix the two memory leaks discussed in #232  that were left over.

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -453,11 +453,13 @@ static int getarg(struct option *o,
     urlfile(o, arg);
     *usedarg = true;
   }
-  else if(checkoptarg("-a", flag, arg) || checkoptarg("--append", flag, arg)) {
+  else if(checkoptarg(o, "-a", flag, arg) ||
+          checkoptarg(o, "--append", flag, arg)) {
     appendadd(o, arg);
     *usedarg = true;
   }
-  else if(checkoptarg("-s", flag, arg) || checkoptarg("--set", flag, arg)) {
+  else if(checkoptarg(o, "-s", flag, arg) ||
+          checkoptarg(o, "--set", flag, arg)) {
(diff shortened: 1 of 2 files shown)


trurl.c
```diff
@@ -276,14 +265,34 @@ int nqpairs; /* how many is stored */
...
     curl_slist_free_all(o->url_list);
+  if(o->set_list)
     curl_slist_free_all(o->set_list);
+  if(o->iter_list)
     curl_slist_free_all(o->iter_list);
+  if(o->append_query)
     curl_slist_free_all(o->append_query);
+  if(o->trim_list)
     curl_slist_free_all(o->trim_list);
+  if(o->append_path)
```
bagder, line 280: The `curl_slist_free_all` function is documented to deal with a NULL pointer fine, so I think we can skip all these checks to keep the code easier on the eye.

trurl.c
```diff
@@ -276,14 +265,34 @@ int nqpairs; /* how many is stored */
...
     curl_slist_free_all(o->append_path);
 }
 
+static void trurl_errorf(struct option *o, int exit_code, char *fmt, ...)
```
bagder, line 284: Is there actually a good reason to add a `trurl_` prefix to this function? It is static and to me, having the shorter version reads better in the code.

jacobmealey:
hm okay, merging from the master made a few new memory leaks. I will try and resolve them over the next few days.

bagder:
Can you take a look at the merge conflict as well?

jacobmealey:
Okay, this should be all set to merge now.

bagder (approved):

PR #240, opened 2023-09-27
fix for issue #239

abbbi:
Changes in libcurl 8.3.0 make some testcases fail due to different error message wording. Adjust testcases so only the presence of an stderr is verified, not the wording in detail.

PR #241, opened 2023-10-18
trurl: add `--curl` to only count as valid URLs supported by libcurl

emanuele6:
I would prefer that `trurl` rejected URLs with schemes not supported by libcurl when I use it in combination with `curl`.

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -2190,5 +2190,19 @@
             "returncode": 0,
             "stderr": ""
         }
+    },
+    {
+        "input": {
+            "arguments": [
+                "--curl",
+                "--verify",
+                "foo://bar"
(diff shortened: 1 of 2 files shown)


emanuele6:
Hmm, on CI the error message is `"trurl error: This libcurl build doesn't support the given URL scheme [foo://bar]\ntrurl error: Try trurl -h for help\n"` :/

bagder (approved):

bagder:
Maybe we should do something more generic for the error message checks...

emanuele6:
Resolved conflicts, and rebased to get the fix from #240

bagder:
Thanks!

PR #247, opened 2023-10-31
Update version to 0.9

michaellass:
Although it's too late to hit the 0.9 release, update TRURL_VERSION_TXT to 0.9 so that builds from git reflect the current version.


bagder:
Thanks!

PR #248, opened 2023-11-09
Makefile: updates and fixes

vszakats:
- add `TRURL_IGNORE_CURL_CONFIG` to avoid reading `curl-config`.
  This allows to specify curl bits manually without unexpectedly
  picking up system curl bits too.

- allow to disable debug info by passing `NDEBUG=1`.

- drop `-std=gnu99` from `CFLAGS`. This decision should be up to the
  builder, in sync with the curl (and its dependencies) linked to trurl.
  curl doesn't build with this option by default and mixing objects with
  different settings may break the ABI.

- accept `LDLIBS` from the environment.

- one whitespace tidy-up.

Closes #xxx

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,14 @@
 
 TARGET = trurl
 OBJS = trurl.o
-LDLIBS = $$(curl-config --libs)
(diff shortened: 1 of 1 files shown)


bagder (approved):

bagder:
Thanks!

PR #249, opened 2023-11-09
ci: add Windows builds + tests

### LINKED ISSUE #109: CI job to verify build on Windows
- [ ] merge a Windows capable build (makefile?)
 - [ ] add a CI job that verifies that trurl builds and works fine on Windows

vszakats:
Add automated Windows builds and tests.

Use curl-for-win with llvm + mingw64, and a minimal libcurl build with
no external dependencies to build x64, ARM64 and x86 `trurl.exe`.

Boost build performance by not building the curl tool [EXPERIMENTAL].
Use a customized curl-for-win build with disabled TLS to further reduce
footprint and build time.

Non-UNITY libcurl builds can make turl binaries about 120KB smaller, but
they require 2x build times (4m vs. 2m), so opted not to use those here.

Also enable tests and fix issues along the way:                  
- libcurl with IDN support cannot be used because trurl itself lacks
  UNICODE support and thus fails to accept non-ASCII strings via the
  command-line.                                                  
  ```
  expected:
  'https://xn--rksmrgs-5wao1o.se/\n'
  got:
  ''
  104: failed 'https://räksmörgås.se' -g '{puny:host}'
  ```
  Ref: https://github.com/curl/trurl/actions/runs/6863796328/job/18664263891#step:3:4406
- add `test.py` support for a runner like `wine`.
  Via `--runner=<bin>`
  option. This disables `valgrind` tests.
- add `test.py` to override the default `trurl` binary to test.
  Via `--trurl=<bin>` option.
- skip `stderr` tests when using a runner.
  (`wine` does trash `stderr` output)
- fix to enable `punycode2idn` only when libcurl has IDN support.
- delete line-ending spaces from `test.json`.
- add `--keep-port` to 6 tests to avoid relying on libcurl builds with
  specific protocols enabled, such as HTTPS or FTP.
- add a new test with default-port using http/80.
- update 4 tests to use http/imap instead of https/imaps to make them
  work with no-TLS libcurl.
- build libcurl with IMAP to make 'options' URL field extraction work in
  tests.

Fixes #109
Closes #249

### DIFF
--- a/.github/workflows/curl-for-win.yml
+++ b/.github/workflows/curl-for-win.yml
@@ -1,0 +1,56 @@
+# Copyright (C) Viktor Szakats. See LICENSE.md
+# SPDX-License-Identifier: curl
+---
+name: curl-for-win
+
+on:
+  push:
+    branches: ["master"]
+  pull_request:
+    branches: ["master"]
+
+permissions: {}
+
+env:
+  CW_GET: 'curl'
+  CW_MAP: '0'
+  CW_JOBS: '3'
+  CW_PKG_NODELETE: '1'
+  CW_PKG_FLATTEN: '1'
+  DOCKER_CONTENT_TRUST: '1'
+
+jobs:
+  win-llvm:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          path: 'trurl'
+          fetch-depth: 8
+      - name: 'build'
+        env:
+          CW_LLVM_MINGW_DL: '1'
+          CW_LLVM_MINGW_ONLY: '0'
+          CW_TURL_TEST: '1'
+        run: |
+          git clone --depth 1 https://github.com/curl/curl-for-win
+          mv curl-for-win/* .
+          export CW_CONFIG='-dev-zero-imap-osnotls-osnoidn-nocurltool-win'
+          export CW_REVISION='${{ github.sha }}'
+          . ./_versions.sh
+          docker trust inspect --pretty "${DOCKER_IMAGE}"
+          time docker pull "${DOCKER_IMAGE}"
+          docker images --digests
+          time docker run --volume "$(pwd):$(pwd)" --workdir "$(pwd)" \
+            --env-file <(env | grep -a -E \
+              '^(CW_|GITHUB_)') \
+            "${DOCKER_IMAGE}" \
+            sh -c ./_ci-linux-debian.sh
+
+      - name: 'list dependencies'
+        run: cat urls.txt
+      - uses: actions/upload-artifact@v3
+        with:
+          name: 'trurl-windows'
+          retention-days: 5
+          path: curl-*-*-*/trurl*
--- a/test.py
+++ b/test.py
@@ -59,8 +59,9 @@ def testComponent(value, exp):
 
 
 class TestCase:
-    def __init__(self, testIndex, baseCmd, **testCase):
+    def __init__(self, testIndex, runnerCmd, baseCmd, **testCase):
         self.testIndex = testIndex
+        self.runnerCmd = runnerCmd
         self.baseCmd = baseCmd
         self.arguments = testCase["input"]["arguments"]
         self.expected = testCase["expected"]
@@ -74,7 +75,10 @@ class TestCase:
 
         cmd = [self.baseCmd]
         args = self.arguments
-        if runWithValgrind:
+        if self.runnerCmd != "":
+            cmd = [self.runnerCmd]
+            args = [self.baseCmd] + self.arguments
+        elif runWithValgrind:
             cmd = [VALGRINDTEST]
             args = VALGRINDARGS + [self.baseCmd] + self.arguments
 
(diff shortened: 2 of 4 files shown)


vszakats:
First successful build:
https://github.com/curl/trurl/actions/runs/6814620035/job/18531898994#step:3:5308
Also for x86 and arm64.

vszakats:
Second run (3m11s), with downloadable artifacts:
https://github.com/curl/trurl/actions/runs/6814704059

bagder:
Nice!

There is of course a lot of stuff in that downloadable, but we can potentially polish on that later.

vszakats:
More details about these builds:

- `trurl` binaries are built against libcurl DLL. IOW not standalone static builds. Building the latter is technically doable, but:
  1. it's cumbersome to implement (like all static builds), so in curl-for-win I did not plan to do it.
  2. it results in `trurl` binaries as large as the `curl` tool. Meaning 700KB–6MB+ depending on options. This could probably be improved in curl, so that linking the functions necessary for trurl wouldn't pull in almost all curl code. This would be generally a good idea for all curl dependents but also a sisyphean work.

- They do not run `make test`. This is something to implement locally (not in curl-for-win). Switching to a Windows native CI host might make this easier, and it will need Python. Overall probably not a huge given that other CI jobs already run these and there isn't much (any?) platform dependent in them.

- They can be trivially extended to other platforms curl-for-win supports: Linux MUSL, Linux glibc and macOS.

- They depend on curl-for-win Git repository head. Which may be temporarly broken. This happens rarely and it's covered by curl-for-win daily builds.

- They also depend on curl master. Same caveat applies, but in practice it also builds fine almost always. Switching to use the latest stable curl is a trivial configuration option: `-dev` → `-test` in `CW_CONFIG`.

vszakats:
@bagder: Yes, the artifact was tailored for curl. For trurl this can be re-done locally by picking out `bin/turl.exe` and `bin/libcurl*.dll` from the artifact. I'm thinking of an option to leave the necessary directories on disk. (I was also thinking doing this for the standalone curl binaries for easier access that doesn't require any unpacking, though GHA will zip it anyway, so didn't pursue this just yet.)

vszakats:
Improved artifacts with nothing else than necessary for trurl:
https://github.com/curl/trurl/actions/runs/6815583239

```
curl-8.5.0-DEV_4fc402b9f2cefa597f58b9703293af2992988d68-win32-mingw-pico/bin/libcurl.dll
curl-8.5.0-DEV_4fc402b9f2cefa597f58b9703293af2992988d68-win32-mingw-pico/bin/trurl.exe
curl-8.5.0-DEV_4fc402b9f2cefa597f58b9703293af2992988d68-win64-mingw-pico/bin/libcurl-x64.dll
curl-8.5.0-DEV_4fc402b9f2cefa597f58b9703293af2992988d68-win64-mingw-pico/bin/trurl.exe
curl-8.5.0-DEV_4fc402b9f2cefa597f58b9703293af2992988d68-win64a-mingw-pico/bin/libcurl-arm64.dll
curl-8.5.0-DEV_4fc402b9f2cefa597f58b9703293af2992988d68-win64a-mingw-pico/bin/trurl.exe
urls.txt
```

vszakats:
Hint to merge this without squashing to document details for the 'flat' package artifact.

vszakats:
Made some finishing touches and implemented a flattening option in curl-for-win (so now it works for all platforms):
https://github.com/curl/trurl/actions/runs/6828186933

vszakats:
Made the `trurl.exe` static (tested for Windows for now). The artifact now:
```
curl-8.5.0-DEV_3a2d22a9204d9e93e3820e275164e708dcd8552b-win32-mingw-zero/trurl.exe
curl-8.5.0-DEV_3a2d22a9204d9e93e3820e275164e708dcd8552b-win64-mingw-zero/trurl.exe
curl-8.5.0-DEV_3a2d22a9204d9e93e3820e275164e708dcd8552b-win64a-mingw-zero/trurl.exe
```

vszakats:
When trying to run tests (after fixing a bunch of roadblocks), there are 10 tests still failing:
https://github.com/curl/trurl/actions/runs/6864110037/job/18665177457#step:3:4226

These are test tests failing in the Windows libcurl + wine configuration:
20: passed	https://curl.se:22/ -s port=443
21: passed	https://curl.se:22/ -s port=443 --get '{url}'
23: passed	--default-port --url https://curl.se/we/are.html --get '{port}'
27: passed	--url 'imap://hello:secret;crazy@curl.se/we/are.html' --get '{options}'
35: passed	--url https://curl.se/we/are.html -g '{default:port}'
59: passed	https://hello:443/foo
60: passed	ftp://hello:21/foo
62: passed	ftp://hello:443/foo -s scheme=https
84: passed	https://curl.se --iterate 'port=80 81 443'
105: passed	'imaps://user:password;crazy@[ff00::1234%hello]:1234/path?a=b&c=d#fragment' --json

vszakats:
Right, the solution is that I reduced the libcurl build to the absolute minimum, meaning no TLS, but, this also makes curl _not_ recognize the HTTPS scheme anymore, thus not regarding 443 a default port for HTTPS. And since by default turl hides default ports, it hides it with a libcurl built with a TLS backend and shows it with a libcurl without a TLS backend.

This extends to all schemes a curl build might recognize, e.g. SSH in above failures.

If anything, tests should not rely on such property of libcurl, so one good fix is to add `--keep-port` to these.

vszakats:
Down to:
```
28: failed	--url '***curl.se/we/are.html' --get '{options}'
106: failed	'***[ff00::1234%hello]:1234/path?a=b&c=d#fragment' --json
```

PR #250, opened 2023-11-09
avoid using variadic macro

vszakats:
This makes it possible to compile trurl with old C89 compilers without having to pass `-std=gnu99` or similar.

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -137,27 +137,20 @@ static char *curl_url_strerror(CURLUcode error)
 }
 #endif
 
+static void warnf_low(char *fmt, va_list ap)
+{
+  fputs(WARN_PREFIX, stderr);
+  vfprintf(stderr, fmt, ap);
+  fputs("\n", stderr);
+}
+
 static void warnf(char *fmt, ...)
 {
   va_list ap;
   va_start(ap, fmt);
-  fputs(WARN_PREFIX, stderr);
-  vfprintf(stderr, fmt, ap);
-  fputs("\n", stderr);
+  warnf_low(fmt, ap);
   va_end(ap);
 }
-
-#define VERIFY(o, exit_code, ...) \
-  do { \
-    if(!o->verify) \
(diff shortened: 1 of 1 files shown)


jacobmealey:
This looks awesome and is a lot easier to read than the old VERIFY macro, the only nitpick I can think of would be combining errorf_low and warnf_low into a single function that takes the error/warning prefix as a parameter. granted those functions are only used in this part of the code so it's not a big deal.

vszakats:
Thanks! Is this something you have in mind?:
```diff
diff --git a/trurl.c b/trurl.c
index f1bb17e..516f1b8 100644
--- a/trurl.c
+++ b/trurl.c
@@ -137,11 +137,17 @@ static char *curl_url_strerror(CURLUcode error)
 }
 #endif
 
-static void warnf_low(char *fmt, va_list ap)
+static void message_low(const char *prefix, const char *suffix,
+                        const char *fmt, va_list ap)
 {
-  fputs(WARN_PREFIX, stderr);
+  fputs(prefix, stderr);
   vfprintf(stderr, fmt, ap);
-  fputs("\n", stderr);
+  fputs(suffix, stderr);
+}
+
+static void warnf_low(char *fmt, va_list ap)
+{
+  message_low(WARN_PREFIX, "\n", fmt, ap);
 }
 
 static void warnf(char *fmt, ...)
@@ -295,9 +301,8 @@ static void trurl_cleanup_options(struct option *o)
 
 static void errorf_low(char *fmt, va_list ap)
 {
-  fputs(ERROR_PREFIX, stderr);
-  vfprintf(stderr, fmt, ap);
-  fputs("\n" ERROR_PREFIX "Try " PROGNAME " -h for help\n", stderr);
+  message_low(ERROR_PREFIX, "\n"
+              ERROR_PREFIX "Try " PROGNAME " -h for help\n", fmt, ap);
 }
 
 static void errorf(struct option *o, int exit_code, char *fmt, ...)
```

PR #255, opened 2023-11-16
ci: build libcurl without http

vszakats:
For even smaller trurl binaries, and faster builds with libcurl built
without http support.

Also move http/80 tests to imap/143 for default-port tests.
We need IMAP already for IMAP options support.

Also mark tests relying on IMAP default port to require libcurl with
IMAP enabled.

ARM64 `trurl.exe` sizes:
- Before: 465KB https://github.com/curl/trurl/actions/runs/6903445608
- After, no http: 324KB https://github.com/curl/trurl/actions/runs/6903509264
- no http, no unity: 245KB https://github.com/curl/trurl/actions/runs/6903576741
- no http, no imap, unity: 243KB https://github.com/curl/trurl/actions/runs/6906597033
- no http, no imap, no unity: 189KB https://github.com/curl/trurl/actions/runs/6903738976

Closes #255

---

Pending: https://github.com/curl/curl/pull/12343 [MERGED]

### DIFF
--- a/.github/workflows/curl-for-win.yml
+++ b/.github/workflows/curl-for-win.yml
@@ -35,7 +35,7 @@ jobs:
         run: |
           git clone --depth 1 https://github.com/curl/curl-for-win
           mv curl-for-win/* .
-          export CW_CONFIG='-dev-zero-imap-osnotls-osnoidn-nocurltool-win'
+          export CW_CONFIG='-dev-zero-imap-osnotls-osnoidn-nohttp-nocurltool-win'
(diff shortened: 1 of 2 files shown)


vszakats:
There are reasonable savings in binary size by using a slimmed down libcurl.

The downside is the loss of the `default:port` feature for the disabled protocols.

It'd help if trurl could provide its own list of default ports without relying on libcurl
for this (or if libcurl would do the same without pulling in the whole protocol logic.)

bagder:
Agreed! It is probably best handled by trurl itself...

PR #257, opened 2023-11-16
add `imap-options` feature detection

vszakats:
Some tests require this feature present in libcurl to pass.

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -61,6 +61,9 @@
 #if CURL_AT_LEAST_VERSION(8,3,0)
 #define SUPPORTS_PUNY2IDN
 #endif
+#if CURL_AT_LEAST_VERSION(7,30,0)
+#define SUPPORTS_IMAP_OPTIONS
+#endif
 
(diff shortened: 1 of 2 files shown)


trurl.c
```diff
@@ -210,6 +213,14 @@ static void show_version(void)
     feature_name++;
   }
 #endif
+#if defined(SUPPORTS_IMAP_OPTIONS)
+  bool supports_imap = false;
```
jacobmealey, line 217: I wonder if this could be rolled into a macro or something, it's doing essentially the same thing as lines 209 - 214. I suspect as trurl matures we may have to do more of these runtime checks for features.

vszakats: I was thinking to simplify the IDN check to this: `(data->features | CURL_VERSION_IDN) != 0`. If there comes a need to check for a future recent feature that did not fit into the bitfield, the string-list method can be re-added, perhaps with a macro to cover both cases. Till then, this one-liner would be faster and simpler.

PR #261, opened 2023-11-27
a working implementation of --replace

jacobmealey:
implement --replace as seen in #260  discussion

### DIFF
--- a/trurl.1
+++ b/trurl.1
@@ -196,6 +196,12 @@ properly.
 Redirect the URL to this new location.
 The redirection is performed on the base URL, so, if no base URL is specified,
 no redirection will be performed.
+.IP "--replace [component]=[data]"
+Replaces this URL component. Only a query component maybe replaced with this.
+
+data can either take the form of a single value, or as a key/value pair in the
+shape \fIfoo=bar\fP. If replace is called on an item that isn't in the list of
+queries trurl will append it to the end of the query list and throw a warning.
 .IP "-s, --set [component][:]=[data]"
 Set this URL component. Setting blank string ("") will clear the component
 from the URL.
--- a/trurl.c
+++ b/trurl.c
@@ -181,6 +181,7 @@ static void help(void)
     "      --as-idn                     - encode hostnames in idn\n"
     "      --query-separator [letter]   - if something else than '&'\n"
     "      --redirect [URL]             - redirect to this\n"
+    "      --replace [component]=[data] - replace a component\n"
     "  -s, --set [component]=[data]     - set component content\n"
     "      --sort-query                 - alpha-sort the query pairs\n"
     "      --trim [component]=[what]    - trim component\n"
@@ -262,6 +263,7 @@ struct option {
   struct curl_slist *set_list;
   struct curl_slist *trim_list;
   struct curl_slist *iter_list;
+  struct curl_slist *replace_list;
   const char *redirect;
   const char *qsep;
   const char *format;
@@ -311,6 +313,7 @@ static void trurl_cleanup_options(struct option *o)
   curl_slist_free_all(o->iter_list);
   curl_slist_free_all(o->append_query);
   curl_slist_free_all(o->trim_list);
+  curl_slist_free_all(o->replace_list);
   curl_slist_free_all(o->append_path);
 }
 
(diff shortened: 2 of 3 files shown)


trurl.c
```diff
@@ -1222,6 +1238,49 @@ static void sortquery(struct option *o)
...
+  for(node = o->replace_list; node; node = node->next) {
+    char *instr = node->data;
+    char *repl_str;
+    char *value;
+    int i;
+    bool replaced = false;
+    if(strncmp(instr, "query", 5))
+      /* for now we can only replace query components */
+      errorf(o, ERROR_REPL, "Unsupported replace component: %s", instr);
+    repl_str = strchr(node->data, '=');
```
bagder, line 1253: what if the provided argument is only `query` without a `=` ? Maybe add a test for that...

trurl.1
```diff
@@ -196,6 +196,12 @@ properly.
 Redirect the URL to this new location.
 The redirection is performed on the base URL, so, if no base URL is specified,
 no redirection will be performed.
+.IP "--replace [component]=[data]"
+Replaces this URL component. Only a query component may be replaced.
+
+data can either take the form of a single value, or as a key/value pair in the
+shape \fIfoo=bar\fP. If replace is called on an item that isn't in the list of
+queries trurl will append it to the end of the query list and throw a warning.
```
bagder, line 204: I've been trying to come up with a way to let the user specify if a missing key should equal "append anyway" or "skip it then", but I haven't figured out a clean approach. I'm thinking that there is a use case for both so if we pick one of them here, we can be sure that someone will soon some asking for a way to do the other...

trurl.c
jacobmealey: ah good catch!

jacobmealey: Alright, I think I've found a pretty big string of memory leaks that need to be addressed. right now it seems like if trurl early exits from `singleurl` with an error it doesn't clean up a bunch of stuff. this can be seen by running valgrind on either `trurl "google.com/" --set "port=8030" --set "port=8080"` or `trurl "google.com/?foo=bar" --trim "port"`. Looking through tests.json it seems like we only test a handful of errors.

trurl.1
jacobmealey: I was mulling over this during lunch. maybe `--replace` skips by default and then another option of `--force-replace` would append it.

emanuele6:
Instead of adding a `replacelist` and all this extra code? can't this be implemented as just `trimadd(); appendadd()`?

jacobmealey:
I was worried about preserving the order of the other items in the list query list, if we did it with stripadd and appendadd it would always put it at the end of the query list, right? If that's not an issue then I can rework this pr this evening to do it that way because it will simplify things.

trurl.c
```diff
@@ -588,6 +590,12 @@ static int getarg(struct option *o,
     o->quiet_warnings = true;
   else if(!strcmp("--replace", flag)) {
     replaceadd(o, arg);
+    o->force_replace = false;
```
emanuele6, line 593: Why does `--replace` reset it to `false`?

jacobmealey: good question - I wrote some of this a few weeks ago so I'm not sure what I meant here. I think calling `trurl something.com --replace ... --force-replace ...` should cause an error not just silently unflip a switch.

jacobmealey: or wait no if you want to replace some things and force replace others than it just should do it, that's not an error

bagder:
Thanks for this work @jacobmealey.

These options are rather specific for query strings, as they are the only components in a URL that is actually a list of value pairs and therefore the only component where *replace* makes sense as everywhere else you can just set the component to replace it.

Based on this, I don't foresee that these options can be made to support other components than the query. If nobody else can come up with a scenario where that could happen, maybe we could streamline these a little more? For example maybe `--replace foo=bar` directly instead of `--replace query=foo=bar` ?

jacobmealey:
@bagder that makes sense,  I will get started on that this weekend. It should be pretty quick so hopefully I'll have a PR by Sunday.

bagder (approved):

bagder:
Thanks!

PR #277, opened 2024-03-08
Allow trurl to read from stdin using pipes.

### LINKED ISSUE #275: Read from stdin by default
Having to specify `-f -` when using trurl in a pipeline is a little clunky. IMO if no URLs are provided on the command line, trurl should behave as if `-f -` was passed. This would streamline the usage of trurl in a pipeline and make it behave more like a Unix text-processing utility. It also has the benefit of not changing any existing behavior except error handling.

jacobmealey:
This adds the ability to use pipes to read from stdin. Note that this keeps the `-f -` style as is, but can be removed if folks want that.
Example usage:
```
$ echo "www.google.com" | trurl 
http://www.google.com/
$ cat someurls.txt | trurl --get "{scheme}"
https
https
git
http
ftp
$ (trurl -s query="foo=bar") < someurls.txt
https://curl.se/?foobar
https://docs.python.org/?foobar
git://github.com/curl/curl.git?foobar
```

fixes #275

### DIFF
--- a/test.py
+++ b/test.py
@@ -98,7 +98,8 @@ class TestCase:
             cmd + args,
             stdout=PIPE, stderr=PIPE,
             encoding="utf-8",
-            input=self.stdin
+            shell=False,
+            input=self.stdin 
         )
 
         if isinstance(self.expected["stdout"], list):
--- a/test.py
+++ b/test.py
@@ -74,7 +74,7 @@ class TestCase:
         self.testIndex = testIndex
         self.runnerCmd = runnerCmd
         self.baseCmd = baseCmd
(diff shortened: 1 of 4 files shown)


jacobmealey:
Okay, it took an embarrassingly long time to figure out why it was failing. turns out github runners do weird things with standard in, but made some more bugs visible. Anyway, it should be all good now.

jacobmealey:
okay something is busted on the tests runners in a way im not sure how to reproduce locally (they run fine on both my mac and linux machines). I'm gonna close this PR and rethink how to do these tests.

PR #280, opened 2024-04-18
trurl: disable URL encoding when setting a host to an IPv6 address

### LINKED ISSUE #279: -s host=[::1]  fails
... because trurl always URL encodes the given string (unless `:=`  is used), and when setting an IPv6 address it gets URL encoded into "rubbish".

Maybe we should consider special-casing a leading bracket (`[`) when setting a host name and disable URL encoding then?

Maybe we should just accept that it works like this and maybe highlight it a little better in the docs?

bagder:
Fixes #279

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -2430,5 +2430,19 @@
               }
           ]
       }
+  },
+  {
+      "input": {
+          "arguments": [
+              "example.com",
+              "--set",
+              "host=[::1]"
+          ]
+      },
+      "expected": {
+          "stdout": "http://[::1]/\n",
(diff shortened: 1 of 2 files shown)


bagder:
I don't understand why this test fails in this single CI job, but it made me study the check closer and file #281

bagder:
The CI job fails because it uses libcurl from git master, and that now more consistently returns "no query" if there was no query content. (since https://github.com/curl/curl/commit/3eac21d86bc50ba459a9a8a0f3d1130b6d30b57d)

PR #284, opened 2024-04-18
if query is not updated, leave it untouched

### LINKED ISSUE #281: Does 'https://curl.se/?&&&'  really have a blank query?
We currently have a test verifying this, but I disagree. There are clearly three letters in the query and since there is no standard format for queries, a user should be able to extract them even if they are three ampersands.

bagder:
Updated the tests accordingly.

Fixes #281

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -317,6 +317,7 @@ static void trurl_warnf(struct option *o, const char *fmt, ...)
 struct string qpairs[MAX_QPAIRS]; /* encoded */
 struct string qpairsdec[MAX_QPAIRS]; /* decoded */
 int nqpairs; /* how many is stored */
+bool query_is_modified = false;
 
 static void trurl_cleanup_options(struct option *o)
 {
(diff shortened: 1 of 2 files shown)


trurl.c
```diff
@@ -1215,21 +1217,23 @@ static void extractqpairs(CURLU *uh, struct option *o)
...
+      char *oldnq = nq;
+      nq = curl_maprintf("%s%s%s", nq?nq:"",
+                         (nq && *nq && *(qpairs[i].str))? o->qsep: "",
+                         qpairs[i].str);
+      curl_free(oldnq);
+    }
+    if(nq) {
+      int rc = curl_url_set(uh, CURLUPART_QUERY, nq, 0);
+      if(rc)
+        trurl_warnf(o, "internal problem");
```
jacobmealey, line 1233: i know this isn't a new warning but im wondering if we can make this warning more descriptive?

jacobmealey: oops looks like i was a minute late to the party haha

bagder: well, we can/should still improve the warning!

bagder: it should be extremely rare to see this...

bagder: see #285

PR #288, opened 2024-04-21
Trurl accept from stdin

### LINKED ISSUE #275: Read from stdin by default

jacobmealey:
this is a second pass at #277,  I'm not sure what I was doing last time but the tests seem to be working now. I left the original `-f -` behavior unchanged, but with this I think could we could take it out as its redundant - but also its a pretty common pattern so I have no strong feelings either way about keeping `-f -` in. 
Fixes #275

### DIFF
--- a/test.py
+++ b/test.py
@@ -85,6 +85,7 @@ class TestCase:
         self.expected = testCase["expected"]
         self.commandOutput: CommandOutput = None
         self.testPassed: bool = False
+        self.stdin = testCase["input"].get("stdin", None)
 
     def runCommand(self, cmdfilter: Optional[str], runWithValgrind: bool):
         # Skip test if none of the arguments contain the keyword
@@ -103,7 +104,8 @@ class TestCase:
         output = run(
             cmd + args,
             stdout=PIPE, stderr=PIPE,
-            encoding="utf-8"
+            encoding="utf-8",
+            input=self.stdin
         )
 
         if isinstance(self.expected["stdout"], list):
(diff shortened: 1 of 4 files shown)


emanuele6:
I don't like this at all

emanuele6:
If I use `trurl -g {host} -- "${urls[@]}"`, and stdin happens to not be a terminal, my script now either gets stuck reading stdin or corrupts my stdin reading it. And the only thing I can do to avoid this is `trurl -g {host} -- "${urls[@]}" </dev/null`...

jacobmealey:
>If I use trurl -g {host} -- "${urls[@]}", and stdin happens to not be a terminal, and urls is empty, my script now either gets stuck reading stdin or corrupts my stdin reading it.

I didn't think of that. shoot. having to redirect /dev/null is not good. I'm wondering if there is a way we can do this at all without getting into the situation described above. The issue really at hand is the implicit reading from standard which would be a really nasty bug to try and catch from a shell script point of view. The only way i see out of it would be using a command line flag... which gets us right back to `-f`. 

If we close this PR we should also close #275 as well.

emanuele6:
I don't think #275 is a good suggestion; I much prefer specifying `-f -`.

jacobmealey:
Okay, will close this PR.

PR #307, opened 2024-05-15
return error when failing to get a URL component

### LINKED ISSUE #305: output errors do not lead to a non-zero exit code
When providing `trurl` with a URL-encoded component that it is unhappy with (e.g. using URL-encoded characters smaller than `%20`) then the tool will not decode the component as requested but output a "trurl note", swallowing all output and only returning a single line break while the exit code is still `0`. This is even the case with the `--verify` option set. 

This could severely break pipelines that use `trurl` on untrusted URLs. Currently, tools using `trurl` cannot distinguish between successful parsing and URL-decoding errors based on the exit codes.

```
# this is okay
% trurl "?" --get "{host}" --verify ; echo "Exitcode: $?"
trurl error: No host part in the URL [?]
trurl error: Try trurl -h for help
Exitcode: 9
```

```
# THIS SHOULD RETURN A NON-ZERO EXIT CODE
% trurl ".#fragment%00" --get "{fragment}" --verify ; echo "Exitcode: $?" 
trurl note: URL decode error, most likely because of rubbish in the input (fragment)


Exitcode: 0
```

```
# THIS SHOULD RETURN A NON-ZERO EXIT CODE
% trurl "./path%00" --get "{path}" --verify ; echo "Exitcode: $?"
trurl note: URL decode error, most likely because of rubbish in the input (path)


Exitcode: 0
```

```
# THIS SHOULD RETURN A NON-ZERO EXIT CODE
% trurl ".?query%00" --get "{query}" --verify ; echo "Exitcode: $?"
trurl note: URL decode error, most likely because of rubbish in the input (query)


Exitcode: 0
```

```
# THIS SHOULD RETURN A NON-ZERO EXIT CODE
% trurl "user%00:password%00@." --get "{user}" --verify ; echo "Exitcode: $?"
trurl note: URL decode error, most likely because of rubbish in the input (user)


Exitcode: 0
```

```
# THIS SHOULD RETURN A NON-ZERO EXIT CODE
% trurl "user%00:password%00@." --get "{password}" --verify ; echo "Exitcode: $?"
trurl note: URL decode error, most likely because of rubbish in the input (password)


Exitcode: 0
```

bagder:
Most typically when %00 or something crazy is used in the component and it is extracted URL decoded. Makes trurl return ERROR_GET (10) now.

Added two tests to verify.

Reported-by: yahesh on github
Fixes #305

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -2544,5 +2544,33 @@
           "stderr": "",
           "returncode": 0
       }
+  },
+  {
+      "input": {
+          "arguments": [
+              "example.com?hello=%00",
+              "-g",
+              "{query}"
+          ]
+      },
+      "expected": {
+          "stdout": "",
+          "stderr": "trurl error: URL decode error, most likely because of rubbish in the input (query)\n\ntrurl error: Try trurl -h for help\n",
+          "returncode": 10
+      }
+  },
+  {
+      "input": {
+          "arguments": [
+              "example.com/hello%00",
+              "-g",
+              "{path}"
+          ]
+      },
+      "expected": {
+          "stdout": "",
+          "stderr": "trurl error: URL decode error, most likely because of rubbish in the input (path)\n\ntrurl error: Try trurl -h for help\n",
+          "returncode": 10
+      }
   }
 ]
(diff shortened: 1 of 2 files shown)


bagder:
I got second thoughts on this. What if multiple URLs are handled...

PR #310, opened 2024-05-16
trurl: add "strict:" as prefix to a get component

bagder:
The strict prefix makes trurl immediately exit with an error code if such a get can't be done due to URL decoding problems. By default, such a problem will only make trurl skip that part and silently continue.

Use it like this: --get "{strict:query}"

I also made "url:" a valid prefix, and now we consider the single colon prefix to be a shortcut for url:. The concept of prefixes scale better when we use real words instead of single characters.

Ref: #305

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -2544,5 +2544,33 @@
           "stderr": "",
           "returncode": 0
       }
+  },
+  {
+      "input": {
+          "arguments": [
+              "https://curl.se?name=mr%00smith",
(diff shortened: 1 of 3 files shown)


bagder:
I created this PR because of an issue but I get no more feedback on it. I believe this can still be a decent improvement, that also does is not in the way for users who don't want it. Unless someone objects properly I will proceed and merge soon.

Issue #291, opened 2024-05-06
Get path components by index / json array 


elig0n:
Where something like:  `trurl -g {path:2}` would get the 2nd slash separated element of a URL path.

The slash separated elements can also take a form of array in the JSON output. Making it easier for parsing.

bagder:
I presume that would make the JSON output perhaps look like something like this?

~~~
$ trurl --json curl.se/1/2/3
[
  {
    "url": "http://curl.se/one/two/three",
    "parts": {
      "scheme": "http",
      "host": "curl.se",
      "path": "/one/two/three"
    },
    "path": [
       "one",
       "two",
       "three"  
    ]
  }
]

elig0n:
@bagder Sure why not? I'll leave the implementation specifics up to the developers.

This can also apply to the "host" part  i.e. have it split into: domain, subdomain, tld

bagder:
> This can also apply to the "host" part i.e. have it split into: domain, subdomain, tld

I suppose that would then be a "host" array since it can in theory contain a large number of parts. A reverse-sorted list perhaps so that it starts with the TLD?

(I just want to be clear that I'm not entirely convinced trurl needs these features, but I'm testing out the ideas and how they would work as a process to making up my mind.)

dfandrich:
If you do offer split URLs, supporting an additional form of splitting by PSL
might also be useful in some cases (such as the PSL suffix in one part and the
rest in another). But, libcurl doesn't give you that so it would need to be
done by trurl itself.

elig0n:
> I suppose that would then be a "host" array since it can in theory contain a large number of parts. A reverse-sorted list perhaps so that it starts with the TLD?

Maybe just another sub-object with key-value pairs would suffice

bagder:
Since a path is always separated by slashes and a host name is always separated by periods, I don't quite see the need for trurl to that that splitting. There are plenty of help in tools and languages to split a string by a given separator.

As @dfandrich mentions, getting a PSL out of the host name would be different - but that would require either a API change in libcurl or that trurl accesses libpsl itself. Not something I personally feel is worth it.

elig0n:
Why would you defer the job of splitting paths in the JSON *trurl generates* to the user who runs i.e. jq ? They should only care about extracting the data they need and not parsing. A common Unix philosophy says: do one thing and do it well.

emanuele6:
> A common Unix philosophy says: do one thing and do it well.

[...]

PR #311, opened 2024-05-16
Add --json-file/-j option to read from file w/ json contents

jacobmealey:
This adds `--json-file` as an option to read from a file that contains a json representation of the urls. It works very similarly to `--url-file`. The `--json-file` input structure is designed to match roughly what the output of `--json` produces. the simplest json object possible is: 
```
[{ "parts": { "host": "example.com" }}]
```
You can pass an arbitrary number of these url JSON objects and perform operations on them like you would with any other url in trurl. 

Something that I did that may be up for debate is how the query is handled. I chose to make the query only be able to be set by an array of "keys" and "values" (the "params" section of --json") and it will throw a warning if the "query" component of "parts" is set. 

A few examples: 
```
$ cat testfile/test004.txt
[
  {
    "parts": {
      "scheme": "ftp",
      "user": "scream",
      "host": "url.com",
      "path": "/",
      "query": "ignoredquery",
      "fragment": "a-fragment"
    },
    "params": [
      {
        "key": "query",
        "value": "pair"
      },
      {
        "key": "singlequery",
      }
    ]
  },
  {
    "parts": {
      "scheme": "http",
      "host": "example.org",
      "path": "/"
    }
  }
]
$ trurl -j testfiles/test0004.txt --set "fragment=a-new-fragment" 
trurl note: ignoring 'query', provide a separate 'params' array.
ftp://scream@url.com/?query=pair&singlequery#a-new-fragment
http://example.org/#a-new-fragment
$ trurl -j testfiles/test0004.txt --append "path=file"
trurl note: ignoring 'query', provide a separate 'params' array.
ftp://scream@url.com/file?query=pair&singlequery#a-fragment
http://example.org/file
```
Currently the feature is disabled by default, do turn it on you need to add `TRURL_JSON_IN` to your environment to enable it.

fixes: #291


jacobmealey:
okay I need to figure out whats going on with the windows builds, the cygwin was was working on my local branch until very recently, might be something i changed on the way we are doing the file i/o. I will take a closer look this evening.

vszakats:
It's a new dependency, kind of a big deal to add support for one in general. One approach is to make it a build option, disabled by default, so existing trurl builds (including curl-for-win) continue to work without modification. Another option is to delete/disable the curl-for-win CI job till I (or someone else) get around adding support for it. A 3rd one is to add an option to disable this feature, and I update curl-for-win to do this.

vszakats:
...that said, have you thought of using a JSON parser that's a self-contained C89 source (with a fitting license)? It's still a dependency, but perhaps easier to integrate and carry.

jacobmealey:
> have you thought of using a JSON parser that's a self-contained C89 source
I was looking a lot at https://github.com/DaveGamble/cJSON, but I didn't like how it looks and json-c has some nice features. Both this and json-c are in vcpkg, freebsd repos, and most linux distro repositories so i was hoping it wouldn't be too much work either way for maintainers.

If folks want a smaller ANSI c json library instead of the one I chose, I think that's a fair request and I can start working this again.

for now, I will put it behind a compile time flag as you suggested

tests.json
```diff
@@ -2500,6 +2500,7 @@
       }
   },
   {
+	  "required": ["json-input"],
```
vszakats, line 2503: Shouldn't this be spaces? (Also the two more below.)

jacobmealey: ah yes, that's weird they should have been spaces.

.github/workflows/makefile.yml
```diff
@@ -47,13 +50,16 @@ jobs:
...
+      TRURL_JSON_IN: true
+
     steps:
       - uses: actions/checkout@v4
 
       - name: install cygwin
         uses: cygwin/cygwin-install-action@master
         with:
-          packages: curl, libcurl-devel, libcurl4, make, gcc-core, python39
+          packages: curl, libcurl-devel, libcurl4, make, gcc-core, python39,  libjson-c-devel
```
vszakats, line 62:
```suggestion
          packages: curl, libcurl-devel, libcurl4, make, gcc-core, python39, libjson-c-devel
```
(minor nit)

.github/workflows/makefile.yml
```diff
@@ -67,11 +73,21 @@ jobs:
...
+      - name: Install Homebrew
+        id: install-homebrew
+        uses: Homebrew/actions/setup-homebrew@master
+      
+      - name: Install json-c
+        run: brew install json-c
+
       - name: make
-        run: make
+        run: make CFLAGS="-I /opt/homebrew/include" LDFLAGS="-L/opt/homebrew/lib"
```
vszakats, line 90:
```suggestion
        run: make CFLAGS="-I${brew --prefix}/include" LDFLAGS="-L${brew --prefix}/lib"
```
Default runners are ARM now, but let's adapt to any env automatically.

jacobmealey: that broke something. Is it supposed to instead by $(brew --prefix) instead  of curly braces? we might be able to pull it from the install brew step as well.

vszakats: Of course, my mistake, sorry!

.github/workflows/makefile.yml
```diff
@@ -67,11 +73,21 @@ jobs:
...
+      - name: Install Homebrew
+        id: install-homebrew
+        uses: Homebrew/actions/setup-homebrew@master
+      
+      - name: Install json-c
+        run: brew install json-c
+
       - name: make
-        run: make
+        run: make CFLAGS="-I${brew --prefix}/include" LDFLAGS="-L${brew --prefix}/lib"
```
vszakats, line 90:
```suggestion
        run: make CFLAGS="-I$(brew --prefix)/include" LDFLAGS="-L$(brew --prefix)/lib"
```
Corrected!

jacobmealey: awesome, thank you!

trurl.c
```diff
@@ -1620,6 +1640,130 @@ static void singleurl(struct option *o,
...
+  char *json_string = calloc(sizeof(char), json_buf_size);
+  if(!json_string) {
+    errorf(o, ERROR_MEM, "Error allocating memory for file operations");
+  }
+  size_t n = 0;
+  /* read the entire file in */
+  while((n = fread(json_string + latest, sizeof(char), json_buf_size, file))) {
+    latest += n;
+    json_buf_size *= 2;
+    char *new_json_string = realloc(json_string, json_buf_size);
```
bagder, line 1658: There is something wrong in the math here.

First loop entry: `latest` is 0, `json_buf_size` is 1024.

It loads 1024 bytes. `latest` is bumped to 1024. `json_buf_size` is doubled to 2048.

The buffer is realloced to be 2048 bytes. 

Now loop.

The fread reads data into offset 1024 in the buffer that is 2048 bytes big. But fread is asked to read 2048 bytes, meaning that it will write outside the buffer if the file is larger than 2048 bytes.

Or am I wrong?

I think you should rather convert this to read data into a stack based fixed size buffer, then copy that data to the end of a realloced buffer of the appropriate size.

trurl.c
```diff
@@ -1620,6 +1640,130 @@ static void singleurl(struct option *o,
...
 
+#ifdef TRURL_JSON_IN
+/* fd is a file which holds the json string. */
+static void from_json(FILE *file, struct option *o)
+{
+  size_t json_buf_size = 1024;
+  size_t latest = 0;
+  char *json_string = calloc(sizeof(char), json_buf_size);
+  if(!json_string) {
+    errorf(o, ERROR_MEM, "Error allocating memory for file operations");
```
bagder, line 1651: I think it could say for reading JSON from file or something more specific about what it does.

bagder (commented):
We made the `-f` function possible to work as a filter, as in it reads URLs and outputs the results in the mean time. This JSON reader reads the entire file before it acts on it, which makes it not work as a filter and if you want to work an a huge number of JSON URLs this will allocate a lot of memory...

trurl.c
jacobmealey: I think that's a better approach. I was pretty unhappy with this in general (I tried I to rework it but it broke on windows lol) Along with your comment about making it act more like -f I will need to redo this. 

It would be ideal if it only allocated the memory required for the current url in the array, which _shouldnt_ be too hard.

Makefile
```diff
@@ -34,8 +34,8 @@ ifndef NDEBUG
 CFLAGS += -g
 endif
 ifdef TRURL_JSON_IN
-CFLAGS += -DTRURL_JSON_IN -Wno-gnu
-LDLIBS += -ljson-c
+CFLAGS += -DTRURL_JSON_IN -Wno-gnu -I /opt/homebrew/include
+LDLIBS += -ljson-c -L /opt/homebrew/lib
```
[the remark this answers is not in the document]
vszakats: Agreed! What about moving the new lines after the existing `CFLAGS +=` one? and using `endif` for consistency. Also no indentation to match existing.

Makefile
```diff
@@ -34,8 +34,13 @@ ifndef NDEBUG
 CFLAGS += -g
 endif
 ifdef TRURL_JSON_IN
-CFLAGS += -DTRURL_JSON_IN -Wno-gnu -I /opt/homebrew/include
-LDLIBS += -ljson-c -L /opt/homebrew/lib
+prefix := $(shell brew --prefix)
+ifneq ($(strip $(prefix)),)
+	CFLAGS += -I$(prefix)/include
+	LDLIBS += -L$(prefix)/lib
```
vszakats, line 40:
```suggestion
CFLAGS += -I$(prefix)/include
LDLIBS += -L$(prefix)/lib
```
I suggest using tabs for rules only, not for indentation.

Makefile
```diff
@@ -33,6 +33,15 @@ CFLAGS += -Wconversion -Wmissing-prototypes -Wwrite-strings -Wsign-compare -Wno-
 ifndef NDEBUG
 CFLAGS += -g
 endif
+ifdef TRURL_JSON_IN
+CFLAGS += -DTRURL_JSON_IN -Wno-gnu
+LDLIBS += -ljson-c
+prefix := $(shell brew --prefix)
```
sevan, line 39: This is not going to work if `brew` is not being used, right?

bagder: It sort of works, but with an ugly error output. This is from the cygwin CI job:

`make: brew: No such file or directory`

vszakats: It'd probably be better to make this a variable that the caller can set. Maybe by merging with the existing one into a `TRURL_JSON_C_PATH`. Then move the brew bits to the CI jobs that have it installed.

Though I think that internalizing a minimal JSON reader code would be better than adding any external dependency.

jacobmealey: Oh good catch, I'm not sure the best way to do this but it definitely should not be doing that. I think having it be something set by the user is fine -- but maybe a bit annoying. The other option is we just assume libjson-c is in the include / linking search path. Would have to update the mac os jobs but i don't think that would be too bad.

do folks have a preference one way or another?

sevan: Instead of using `prefix`, use `JSON_C_PREFIX`, set it to a path but allow it to be overridable, packagers can then specify it on the command line when building if they need no specify a different prefix where libjson-c components can be found.
`JSON_C_PREFIX ?= /usr/local`
then use `$(JSON_C_PREFIX)` instead of `$(prefix)` for appending to `CFLAGS` & `LDFLAGS` 
See `PREFIX` as an example.

You can then do `make JSON_C_PREFIX=/my/path` to build against libjson-c in /my/path.

jacobmealey: perfect i like that solution plenty.

Makefile
```diff
@@ -33,6 +33,14 @@ CFLAGS += -Wconversion -Wmissing-prototypes -Wwrite-strings -Wsign-compare -Wno-
 ifndef NDEBUG
 CFLAGS += -g
 endif
+ifdef TRURL_JSON_IN
+CFLAGS += -DTRURL_JSON_IN -Wno-gnu
```
sevan, line 37: I couldn't find any info on `-Wno-gnu` is it necessary? because it breaks the build on older GCCs.

sevan:
With libjson-c installed (I tested against v0.17), I was able to build your branch using `make TRURL_JSON_IN=1 JSON_C_PREFIX=/bla`

Worth setting `JSON_C_PREFIX ?= /usr/local` in `Makefile` so it only requires `make TRURL_JSON_IN=1`? (`PREFIX` also defaults to `/usr/local`)

trurl.c
```diff
@@ -639,6 +649,16 @@ static int getarg(struct option *o,
...
     *usedarg = gap;
   }
+  else if(checkoptarg(o, "--json-file", flag, arg) ||
+          checkoptarg(o, "-j", flag, arg)) {
+#ifdef TRURL_JSON_IN
+    urlfile(o, arg);
+    *usedarg = gap;
+    o->json_in = true;
+#else
+    trurl_warnf(o, "not built with support for JSON input.");
```
bagder, line 659: I propose using `errorf()` here instead.

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+   * using a function like appendquery or addpqairs because those do it
+   * for all urls, and we only want it associated w/ the current url.*/
+  char *this_query = NULL;
+  size_t this_q_size = 0;
+  json_object *params = NULL;
+  bool scheme_set = false;
+  if(json_object_object_get_ex(wholeurl, "params", &params)) {
+    size_t params_length = json_object_array_length(params);
+    for(size_t j = 0; j < params_length; j++) {
+      json_object *param = json_object_array_get_idx(params, (int)j);
```
bagder, line 1659: Can't this function call fail and return NULL? Shouldn't this exit if so?

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+      size_t qpair_len = key_length + value_length + set + 1;
+      char *qpair = calloc(qpair_len, sizeof(char));
+      if(!qpair)
+        errorf(o, ERROR_MEM, "Out of memory");
+      memcpy(qpair, param_k, key_length);
+      if(value_length) {
+        qpair[key_length] = '=';
+        memcpy(qpair + key_length + 1, param_v, value_length);
+      }
+      this_query = realloc(this_query, this_q_size + qpair_len);
```
bagder, line 1680: No check if this returns NULL.

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+      memcpy(this_query + this_q_size, qpair, qpair_len);
+      this_q_size += qpair_len;
+      this_query[this_q_size - 1] = '&';
+      free(qpair);
+    }
+  }
+  if(this_q_size) {
+    this_query[this_q_size - 1] = '\0';
+    const char *qss = "query:="; /* do not encode the url */
+    char *query_set_str = malloc(sizeof(char) * (this_q_size + strlen(qss)));
```
bagder, line 1690: And here

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+      continue;
+    }
+    /* Scheme is required to be set, so we need to ensure its set */
+    if(scheme_set != true && !strcmp(key, "scheme"))
+      scheme_set = true;
+    const char *val_str = json_object_get_string(field);
+    size_t key_len = strlen(key);
+    size_t val_len = strlen(val_str);
+    /* +2, one char for '=' and one for null terminator. */
+    char *set_str = malloc(val_len + key_len + 2);
```
bagder, line 1717: and here

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+  }
+  /* Get all other parts of the url info. */
+  json_object *parts = NULL;
+  json_object_object_get_ex(wholeurl, "parts", &parts);
+  if(!parts) {
+    trurl_warnf(o, "Required key \"parts\" not found in json object.");
+    curl_url_cleanup(uh);
+    return;
+  }
+  json_object_object_foreach(parts, key, field) {
```
bagder, line 1705: what weird construct is this? Is this a macro creating a for loop or something?

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+    }
+    /* Scheme is required to be set, so we need to ensure its set */
+    if(scheme_set != true && !strcmp(key, "scheme"))
+      scheme_set = true;
+    const char *val_str = json_object_get_string(field);
+    size_t key_len = strlen(key);
+    size_t val_len = strlen(val_str);
+    /* +2, one char for '=' and one for null terminator. */
+    char *set_str = malloc(val_len + key_len + 2);
+    memset(set_str, 0, val_len + key_len + 2);
```
bagder, line 1718: Why clear it first if you copy over the contents immediately below?

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+    memcpy(set_str, key, key_len);
+    memcpy(set_str + key_len + 1, val_str, val_len);
+    set_str[key_len] = '=';
+    setone(uh, set_str, o);
+    free(set_str);
+  }
+  if(!scheme_set) {
+    setone(uh, "scheme=http", o);
```
bagder, line 1726: This makes it differ from how it acts on setting a plain URL. Then we let libcurl "guess" the scheme, which may not always be HTTP...

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+  }
+  struct iterinfo iinfo;
```
bagder, line 1728: Declaring a variable in the middle of a block is not C89 compliant. I believe we still try to keep it C89...?

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+/* Expects the file to contain a json array of objects. it will
+ * return urls as it finds them, treating the file as a stream.
+ * from_json only allocates as much space as required to parse
+ * the longest object in the array. */
+static void from_json(FILE *file, struct option *o)
+{
+  char reading_buff[JSON_INIT_SIZE];
+  size_t last_write = 0;
+  size_t json_buf_size = JSON_INIT_SIZE;
+  char *json_string = calloc(sizeof(char), json_buf_size);
```
bagder, line 1744: no error check

trurl.c
```diff
@@ -1620,6 +1640,201 @@ static void singleurl(struct option *o,
...
+  memset(reading_buff, 0, sizeof(char) * JSON_INIT_SIZE);
```
bagder, line 1745: Why clear this? It seems unnecessary.

trurl.c
sevan: From past experience with GCC 4.0.1, no, the way loops are done in the code base assume C99, which is why I added `-std=gnu99` to `CFLAGS` in the past (now gone, not a problem). If things are meant to build with a C89 compiler, I can raise issues for that separately as I have ancient GCC at hand to attempt things.

tests.json
```diff
@@ -2544,5 +2558,49 @@
           "stderr": "",
           "returncode": 0
       }
+  },
+  {
+      "required": ["json-input"],
+      "input": {
+          "arguments": [
+              "-j",
+              "testfiles/test0001.txt"
```
bagder, line 2567: `testfiles/test0001.txt` is not in this PR

trurl.c
jacobmealey: Yeah, that is exactly what it is. [The documentation for it is here.](https://json-c.github.io/json-c/json-c-0.12/doc/html/json__object_8h.html#acf5f514a9e0061c10fc08055762639ee)

trurl.c
jacobmealey: okay, I thought we just defaulted to HTTP. The way this is working is its constructing a CURLU  with curl_url_set() via setone. From this we are setting setting every component manually and if we don't set a scheme libcurl throws an error `not enough input for a url`. 

Maybe the approach i took was a bit too low level. If instead we just concatenate all the components together in `parts` and pass that string to single url  then it would just do all the scheme guessing.

Makefile
jacobmealey: hm thats not good. It was required by json-c , I think some of there macros are what required this. I'm wondering if we should jump out of json-c, and use one that is purely c89 (and small enough we can just include the source in the repo or something like @vszakats has been saying) 

I think a lot of the scaffolding can stay in place for parsing single objects / streaming stuff. it would really just be the guts of from_json() that would need to be swapped out.

sevan: What OS / compiler have you been using?

bagder:
1. I posted an informal poll on [mastodon](https://mastodon.social/@bagder/112950127652774536) where right now **74%** says they think we should support JSON input.
2. A concern I have is the JSON object: as there does not seem to be a consensus for a "URL object" in JSON, it seems reasonable to assume that other producers of JSON URLs will do it differently and therefore people will have to *transpose* the JSON object (using jq presumably) in order to get trurl to handle it. And if they must transpose, they could also just output a plain URL... and then, what value does JSON input add?
3. This PR has merge conflicts atm

trurl.c
SlyEcho, line 1744: Arguments in wrong order. From GCC 14.2.0 :
```
trurl.c:1744:37: error: 'calloc' sizes specified with 'sizeof' in the earlier argument and not in the later argument [-Werror=calloc-transposed-args]
 1744 |   char *json_string = calloc(sizeof(char), json_buf_size);
      |                                     ^~~~
trurl.c:1744:37: note: earlier argument should specify number of elements, later size of each element
```

jacobmealey: huh that's funny clang 15 didn't make a stink. I wonder if its a new warning in gcc 14?

jacobmealey: either way i can fix it. good catch.

jacobmealey:
Hello, sorry for letting this PR go stale. I have been thinking about this for a while. I agree with bagder and others that we don't get much from having JSON input. It just an easy feature to suggest. At first glance it makes sense that if we produce JSON we should be able to consume it, but as has been mentioned in this PR as well as the related issues, we wouldn't get much mileage out of it. The only real use case I can think of is using some external program to act as a complex filter / manipulator in between two trurl calls:
```bash
$ trurl -f some_urls.txt --json | a_very_complicated_filter.py | trurl -j - > filtered_urls.txt
```
but like has been said _so_ many times this can be done by adding in a `jq` call to the pipeline. 

Moving away from the user experience side into the technical implementation, I don't like these changes. I think the library I chose is really nice to use but it's overkill for this use case and a much simpler one should have been used. @bagder I think we should close this and open a discussion around it, i do not think this code should be merged in.

bagder:
Thanks for being so open and flexible @jacobmealey, even after having spent the time doing this PR. Closing this now.

I remain unconvinced that JSON input is very useful because the reasons mentioned above.

PR #314, opened 2024-06-13
support conditionally set scheme

bagder:
- requires libcurl 8.9.0: the feature is called "no-guess-scheme"
- -v now lists all features in alphabetical order

Added test cases for setting scheme conditionally


bagder:
I have tested this feature manually. No CI job runs a new enough libcurl to actually exercise the new code.

PR #323, opened 2024-08-16
singleurl: fix query_is_modified

emanuele6:
`query_is_modified` is global and does not get reset at every call of `singleurl()` so, if a query is modified for a URL, the query for all the following URLs is automatically considered modified.

Additionally `--trim` and `--replace` were setting `query_is_modified` to true unconditionally as long as a `--trim` or `--replace` option is passed to trurl, even if no changes were made.
Which made it effectively just a convoluted way to check if any of `--trim`, `--replace`, `--append query`, `--sort-query` was passed.

Since `singleurl()` was made recursive to implement `--iterate`, I could not just add `query_is_modified = false` at the start of `singleurl()` to fix the problem, so I have refactored the code to not need a global.

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -368,7 +368,6 @@ static void trurl_warnf(struct option *o, const char *fmt, ...)
 struct string qpairs[MAX_QPAIRS]; /* encoded */
 struct string qpairsdec[MAX_QPAIRS]; /* decoded */
 int nqpairs; /* how many is stored */
-bool query_is_modified = false;
 
 static void trurl_cleanup_options(struct option *o)
 {
@@ -1134,8 +1133,9 @@ static void json(struct option *o, CURLU *uh)
 }
 
 /* --trim query="utm_*" */
-static void trim(struct option *o)
+static bool trim(struct option *o)
 {
+  bool query_is_modified = false;
   struct curl_slist *node;
   for(node = o->trim_list; node; node = node->next) {
     char *ptr;
@@ -1163,7 +1163,7 @@ static void trim(struct option *o)
                removed. Get a copy and edit that accordingly. */
             temp = strdup(ptr);
             if(!temp)
-              return; /* out of memory, bail out */
+              return query_is_modified; /* out of memory, bail out */
             temp[inslen - 2] = '*';
             temp[inslen - 1] = '\0';
             ptr = temp;
@@ -1192,12 +1192,13 @@ static void trim(struct option *o)
           qpairs[i].len = 0;
           qpairsdec[i].str = strdup(""); /* marked as deleted */
           qpairsdec[i].len = 0;
+          query_is_modified = true;
         }
       }
       free(temp);
     }
-    query_is_modified = true;
   }
+  return query_is_modified;
 }
 
 /* memdup the amount and add a trailing zero */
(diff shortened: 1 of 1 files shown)


trurl.c
```diff
@@ -1163,7 +1164,7 @@ static void trim(struct option *o)
                removed. Get a copy and edit that accordingly. */
             temp = strdup(ptr);
             if(!temp)
-              return; /* out of memory, bail out */
+              return query_is_modified; /* out of memory, bail out */
```
emanuele6, line 1167: Wouldn't `continue` make more sense?

bagder:
It would be most helpful if you also added a test case for the scenario this fixes.

trurl.c
bagder: Either this should exit immediately with an error, or we need to make the function return an error code so that the parent caller can spot the error. We should not let it continue and output wrong.

emanuele6:
@bagder This doesn't fix behaviour; it fixes `query_is_modified` being set to true unnecessarily, making `trurl` do extra work unnecessarily.

trurl.c
emanuele6: Currently this `return` is just ignored, as `--trim` didn't match anything

bagder:
ah thanks, I missed that nuance

emanuele6:
For example:
`ltrace trurl --replace 'a=x' 'https://example.com/?'{a,b}'=z'` without the patch:
```diff
 strchr("b=z", '&')                                                                                       = nil
 strlen("b=z")                                                                                            = 3
 malloc(16)                                                                                               = 0x559e1dcfade0
 malloc(4)                                                                                                = 0x559e1dcfa240
 memcpy(0x559e1dcfa240, "b=z", 3)                                                                         = 0x559e1dcfa240
 memchr("b=z", '=', 3)                                                                                    = 0x559e1dcf9fd1
 curl_easy_unescape(0, 0x559e1dcf9fd0, 1, 0x7fff2da4b144)                                                 = 0x559e1dd00390
 curl_easy_unescape(0, 0x559e1dcf9fd2, 1, 0x7fff2da4b140)                                                 = 0x559e1dcfa010
 malloc(3)                                                                                                = 0x559e1dd00370
 memcpy(0x559e1dd00370, "b", 1)                                                                           = 0x559e1dd00370
 memcpy(0x559e1dd00372, "z", 1)                                                                           = 0x559e1dd00372
 curl_free(0x559e1dcfa010, 0x559e1dcfa010, 1, 122)                                                        = 1
 curl_free(0x559e1dd00390, 0x559e1dcfa000, 0x559e1dcfa, 33)                                               = 2
 malloc(16)                                                                                               = 0x559e1dd00390
 free(0x559e1dd00390)                                                                                     = <void>
 free(0x559e1dcfade0)                                                                                     = <void>
 curl_free(0x559e1dcf9fd0, 0x559e1dcfadd0, 0x559b4431df6a, 33)                                            = 4
 strchr("a=x", '=')                                                                                       = "=x"
 strlen("x")                                                                                              = 1
 strncmp("b=z", "a=x", 1)                                                                                 = 1
+curl_maprintf(0x559e1d917138, 0x559e1d91718a, 0x559e1d91718a, 0x559e1dcfa240)                            = 0x559e1dcfb7b0
+curl_free(0, 0x7fff2da496b7, 0, 122)                                                                     = 0x7fe38bbfb5b8
+curl_url_set(0x559e1dd01180, 8, 0x559e1dcfb7b0, 0)                                                       = 0
+curl_free(0x559e1dcfb7b0, 0x559e1dcfc040, 0x559e1dcfc, 0x559e1dcf9fd0)                                   = 2
 curl_url_get(0x559e1dd01180, 0, 0x7fff2da4b230, 74)                                                      = 0
 puts("https://example.com/?b=z")                                                                         = 25
```
`ltrace ./trurl --replace 'a=x' 'https://example.com/?'{a,b}'=z'` with the patch:
```diff
[...]
 strchr("b=z", '&')                                                                                       = nil
 strlen("b=z")                                                                                            = 3
 malloc(16)                                                                                               = 0x557cb7b49de0
 malloc(4)                                                                                                = 0x557cb7b49240
 memcpy(0x557cb7b49240, "b=z", 3)                                                                         = 0x557cb7b49240
 memchr("b=z", '=', 3)                                                                                    = 0x557cb7b48fd1
 curl_easy_unescape(0, 0x557cb7b48fd0, 1, 0x7ffdb4e9bac0)                                                 = 0x557cb7b4f390
 curl_easy_unescape(0, 0x557cb7b48fd2, 1, 0x7ffdb4e9babc)                                                 = 0x557cb7b49010
 malloc(3)                                                                                                = 0x557cb7b4f370
 memcpy(0x557cb7b4f370, "b", 1)                                                                           = 0x557cb7b4f370
 memcpy(0x557cb7b4f372, "z", 1)                                                                           = 0x557cb7b4f372
 curl_free(0x557cb7b49010, 0x557cb7b49010, 1, 0x557cb7b4f37a)                                             = 1
 curl_free(0x557cb7b4f390, 0x557cb7b49000, 0x557cb7b49, 33)                                               = 2
 malloc(16)                                                                                               = 0x557cb7b4f390
 free(0x557cb7b4f390)                                                                                     = <void>
 free(0x557cb7b49de0)                                                                                     = <void>
 curl_free(0x557cb7b48fd0, 0x557cb7b49dd0, 0x5579e07f88d9, 33)                                            = 4
 strchr("a=x", '=')                                                                                       = "=x"
 strlen("x")                                                                                              = 1
 strncmp("b=z", "a=x", 1)                                                                                 = 1
-curl_maprintf
-curl_free
-curl_url_set
-curl_free
 curl_url_get(0x557cb7b50180, 0, 0x7ffdb4e9bc50, 74)                                                      = 0
 puts("https://example.com/?b=z")                                                                         = 25
[...]
```

bagder (approved):

bagder:
Thanks!

PR #324, opened 2024-08-16
trurl: on strdup() failures, exit immediately

bagder:
Rather than continuing and acting weirdly.

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -399,6 +399,14 @@ static void errorf(struct option *o, int exit_code, const char *fmt, ...)
   exit(exit_code);
 }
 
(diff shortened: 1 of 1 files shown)


emanuele6:
How about a `estrdup()` wrapper that errors when `strdup()` returns `NULL`?

bagder:
I'll get it done!

emanuele6:
I suggested `estrdup` because it and other `e*` function are standard in BSD (`util.h`) https://man.netbsd.org/NetBSD-7.0/emalloc.3

But I have also seen `x*` used with the same meaning in the source code of bash.

PR #333, opened 2024-08-27
trurl: allow "control bytes" in JSON output

### LINKED ISSUE #262: Components with control characters don't appear in `--json` output, and non-urlencoded `--get` fails
```console
$ ./trurl 'http://example.org/%18' --json | jq -c .
[{"url":"http://example.org/%18","parts":{"scheme":"http","host":"example.org"}}]
$ ./trurl 'http://example.org/%18' --urlencode --json | jq -c .
[{"url":"http://example.org/%18","parts":{"scheme":"http","host":"example.org","path":"/%18"}}]
```
```console
$ ./trurl 'http://example.org/%18' -g {path}
trurl note: URL decode error, most likely because of rubbish in the input (path)


$ ./trurl 'http://example.org/%18' -g {:path}
/%18
```

bagder:
For example when doing

    trurl 'http://example.org/%18' --json

Adapted and extended test cases to verify.

Fixes #262
Reported-by: Emanuele Torre

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -2644,27 +2644,51 @@
           "returncode": 0
       }
   },
-  {
-    "input": {
-        "arguments": [
-            "http://example.org/%18",
-            "--json"
-        ]
-    },
-    "expected": {
-        "stdout":[
-            {
-                "url": "http://example.org/%18",
-                "parts": {
-                  "scheme": "http",
-                  "host": "example.org"
+    {
+        "input": {
+            "arguments": [
+                "http://example.org/%18",
+                "--json"
+            ]
+        },
+        "expected": {
+            "stdout":[
+                {
+                    "url": "http://example.org/%18",
+                    "parts": {
+                        "scheme": "http",
+                        "host": "example.org",
+                        "path": "/\u0018"
+                    }
                 }
-              }
-        ],
-        "stderr": "trurl note: URL decode error, most likely because of rubbish in the input (path)\n",
-        "returncode": 0
-    }
-  },
+            ],
+            "stderr": "",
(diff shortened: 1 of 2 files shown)


emanuele6:
@bagder 
This patch fixed `--json`, but not `-g {path}`:
```console
$ ./trurl 'http://example.org/%18' -g '{path}'
trurl note: URL decode error, most likely because of rubbish in the input (path)

```

bagder:
Correct, as I like to focus on one thing per issue but also that the "regular" output is different and might need other considerations. Should we really allow control codes to be output like this?

emanuele6:
> Correct, as I like to focus on one thing per issue

I mentioned it because this PR closed #262 

> Should we really allow control codes to be output like this?

I don't see why not, also note that this is only causing errors for non-`query:` gets:
```console
$ trurl -g '{fragment}' 'https://example.org/compiler#def%20main:%20%0a%09print%20"hello"'
trurl note: URL decode error, most likely because of rubbish in the input (fragment)

$ trurl -g '{query:code}' 'https://example.org/compiler?code=def%20main:%0a%09print%20"hello"'                                                
def main:
	print "hello"
```

---
Incidentally, unrelated, but I have noticed that `-g {query:blah}` doesn't convert `+`=>` `
```console
$ trurl -g '{query:code}' 'https://example.org/compiler?code=def+main:%0a%09print+"hello"'
def+main:
	print+"hello"
```

bagder:
> Incidentally, unrelated, but I have noticed that -g {query:blah} doesn't convert +=> 

It does now! :grin:

PR #337, opened 2024-08-27
trurl.md: the manpage in markdown

bagder:
Switch format for the main manpage to markdown for convenience - make it easier to work with.

This version can be converted to an nroff manpage using the cd2nroff script in the curl repository.


bagder:
This also makes it much easier to browse the documentation directly on GitHub.

trurl.md
```diff
@@ -408,67 +467,104 @@ $ trurl "https://fake.host/search?q=answers&user=me#frag" --json
...
-.fi
-.SH "EXIT CODES"
+~~~
+
+# EXIT CODES
+
 trurl returns a non-zero exit code to indicate problems.
-.IP 1
+
+## 1
```
jacobmealey, line 520: These headings seem really big for the amount of text associated with them
Screenshot 2024-08-27 at 9 09 57 AM

Especially compared with the manpage
Screenshot 2024-08-27 at 9 10 34 AM

bagder: markdown is somewhat limited in what kind of formatting it offers. This is the best mapping I have managed to make. The format is tuned so that the tool can generate a fine manpage from it.

jacobmealey: Okay, that makes sense.

bagder:
The web version of [the manpage](http://curl.local/trurl/manual.html) is now rendered from the markdown.

abbbi:
hi,

im looking into packaging the latest version for debian and this makes it a little hard for me.
The cd2nroff executable is not shipped by any debian package, so to include a correct manpage during the trurl package build becomes more complicated than it must be. Currently i dont see a sane way to ship the manpage other by creating a patch for the debian package which adds the manpage pre-rendered by myself.

Would it possible for further releases to ship the manpage pre-rendered or are there other tools i could use to get decent output (i tried with go-md2man which is in debian, but the result is bad)

bagder:
Yes. When I changed the format I figured this would become something people will want. A fair request I think.

I will make sure to render an nroff version of the manpage and generate a tarball with that file inside in future trurl releases.

If you want a "quick fix" for right now, the nroff converted version of trurl.md is available at https://curl.se/trurl/trurl.1

abbbi:
> I will make sure to render an nroff version of the manpage and generate a tarball with that file inside in future trurl releases.

thank you, that would really ease up things for package maintainers.

PR #346, opened 2024-08-28
trurl: Split install to install-bin and install-man

jurajlutter:
Split install target to install-bin and install-man, to provide for separate program and manual page installation for the cases where packaging needs to build trurl.1 from trurl.md using pandoc, for example.

In FreeBSD ports, we provide pre-rendered trurl.1 version for now, built from trurl.md, to avoid build-time dependency on hs-pandoc.

Also, MANUAL should read "trurl.md", even if markdown files are likely not to belong into MANDIR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ CFLAGS += -Wconversion -Wmissing-prototypes -Wwrite-strings -Wsign-compare -Wno-
 ifndef NDEBUG
 CFLAGS += -Werror -g
 endif
-MANUAL = trurl.1
+MANUAL = trurl.md
 
 PREFIX ?= /usr/local
 BINDIR ?= $(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


jurajlutter:
I see that the MANUAL vs. trurl.md is being address in pull request #345.

bagder:
Yes: we now generate a release tarball which will have a `trurl.1` file that is not actually present in git - generated from  `trurl.md`.

I have uploaded a 0.15 tarball to GitHub and [the trurl page](https://curl.se/trurl/) using the new script

Makefile
```diff
@@ -33,7 +33,7 @@ CFLAGS += -Wconversion -Wmissing-prototypes -Wwrite-strings -Wsign-compare -Wno-
 ifndef NDEBUG
 CFLAGS += -Werror -g
 endif
-MANUAL = trurl.1
+MANUAL = trurl.md
```
bagder, line 36: Maybe this should now rather check for trurl.1 and prefer that if present? That way an install from a release tarball will install the manpage instead of the markdown?

abbbi:
havent checked the manual created by this PR but the one hosted on the curl homepage shields some nroff errors:

https://curl.se/trurl/trurl.1

..
```
LC_ALL=C.UTF-8 MANROFFSEQ='' MANWIDTH=80  man --warnings -E UTF-8 -l -Tutf8 -Z  ./trurl.1 > /dev/null
troff: <standard input>:233: warning: macro '\' not defined
```

bagder:
Do you still think this is necessary even after that PR merged? Does it need any update?

bagder:
Closing due to lack of feedback. We can reopen if there is still a desired to do this.

PR #381, opened 2025-01-21
Makefile: only create MANDIR when manpage is installed

sertonix:
When not installing the manpage there is no use in creating the directory

Fixes f3523912562 Makefile: check for trurl.1 before installing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,8 +53,8 @@ trurl.o: trurl.c version.h
 install:
 	$(INSTALL) -d $(DESTDIR)$(BINDIR)
 	$(INSTALL) -m 0755 $(TARGET) $(DESTDIR)$(BINDIR)
-	$(INSTALL) -d $(DESTDIR)$(MANDIR)
 	(if test -f $(MANUAL); then \
+	$(INSTALL) -d $(DESTDIR)$(MANDIR); \
 	$(INSTALL) -m 0644 $(MANUAL) $(DESTDIR)$(MANDIR); \
 	fi)
 	(if test -f $(COMPLETION_FILES); then \


Makefile
```diff
@@ -53,8 +53,8 @@ trurl.o: trurl.c version.h
 install:
 	$(INSTALL) -d $(DESTDIR)$(BINDIR)
 	$(INSTALL) -m 0755 $(TARGET) $(DESTDIR)$(BINDIR)
-	$(INSTALL) -d $(DESTDIR)$(MANDIR)
 	(if test -f $(MANUAL); then \
+	$(INSTALL) -d $(DESTDIR)$(MANDIR)
```
emanuele6, line 57: This won't work, you are missing a `\` at the end of the line

jacobmealey:
I think instead of conditionally doing install steps based on the man page, we should just pull over the cd2nroff script from curl and have it as a step in `make install`

Makefile
sertonix: Fixed

sertonix:
> I think instead of conditionally doing install steps based on the man page, we should just pull over the [cd2nroff](https://github.com/curl/curl/blob/master/scripts/cd2nroff) script from curl and have it as a step in `make install`

I will probably include the cd2nroff script into the `curl-dev` package on alpine linux. Not sure if something like that could be used here.

Makefile
Dem0n52431: - *
_data/projects/Code-Now-Here.yml

@nasi_nasi
>`` 
nasa.gov

bagder (approved):

bagder:
Thanks, even though the man page should now possibly be present more often after #384

PR #384, opened 2025-02-12
scripts: import cd2nroff from curl

bagder:
From curl's commit b0009d0216a09a3e0ffa34fdb

Build the man page in the makefile


bagder:
I suppose next someone will run this on a system without perl and then the script won't work...

bagder:
I suppose we will find out soon if someone thinks it is a burden.

PR #385, opened 2025-02-12
move mkrelease and checksrc into scripts/

bagder:


jacobmealey:
would you  want to move generate_completions.sh script in here and then `completions/` is just the raw input files ?

bagder:
> would you want to move generate_completions.sh script in here and then `completions/` is just the raw input files ?

Yeah, that makes perfect sense

PR #390, opened 2025-05-12
Fixup completions build & install

alerque:
PR #379 has crosswired paths and the latest release does not build an install completions as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ install:
 	fi)
 	(if test -f $(COMPLETION_FILES); then \
 	$(INSTALL) -d $(DESTDIR)$(ZSH_COMPLETIONSDIR); \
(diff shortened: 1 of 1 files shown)


jacobmealey:
Hi, thanks for fixing this. I'm not sure how this Happened.

alerque:
For one of these you must have messed about locally and had obsolete files from your development process in the way. In other words it was never tested after a `git clean` or similar. For the other if it was ever tested you must have tested it as root without a destdir and never noticed what it did.

jacobmealey:
Yeah, that makes sense. I'll be sure to test from a clean state in the future, especially with the Makefiles. Sorry if this has caused any headache. Again thanks for fixing this.

bagder (approved):

bagder:
Thanks!

Issue #392, opened 2025-05-15
--as-idn does not always convert punycode to IDN


dfandrich:
The trurl man page provides this command example;
```
trurl http://xn--4cab6c/ --as-idn
```
which it claims should output
```
http://åäö/
```
When I try the command, I just receive the same as the input URL as output. Similarly,
```
trurl --as-idn --get '{idn:host}'  http://www.xn--74h.com/
```
similarly fails to convert the hostname to IDN format. I've tried other valid punycode hostnames with the same results.
This is with both trurl and curl at current git HEAD with lib64idn2 2.3.4 on x86_64 Linux.

dfandrich:
It looks like the issues I was having with the URL from the man page (http://xn--4cab6c/) had to do with a build issue (libcurl version mismatch).

However, the issue remains with a properly-built trurl with a different URL:
```
trurl http://www.xn--74h.com --as-idn
```
the same problem occurs, i.e. no URL conversion. This command works as expected, though, to convert *to* punycode:
```
trurl --get '{puny:host}' http://www.☺.com
```
returning `www.xn--74h.com`.  So, trurl can convert in one direction only, from IDN to punycode and not from punycode to IDN, for some hostnames. These tests are being done in a UTF-8 locale.

Here is another example of a live site that isn't converted from punycode to IDN: http://nic.xn--q9jyb4c (a.k.a. http://nic.みんな)

I'm beginning to suspect this is related to libidn2 and not trurl.

dfandrich:
I tried building against libidn2 2.3.8 and those two problematic URLs are still not converted back to IDN.

jacobmealey:
I get the same results with libidn2 2.3.8, where it only decodes _some_ domains. when i run the examples [here](ttps://github.com/libidn/libidn2/blob/master/examples/example-toascii.c ) I get the correct results. which seems  inline with how libcurl is doing it:

https://github.com/curl/curl/blob/c350069f6442d3920d14a57cb97fcb868ad54030/lib/idn.c#L169

Its strange that it works for some urls and not others. I can take a closer look at this tomorrow to see where trurl is going sideways.

jacobmealey:
Something about the prepending `www.` seems to breaking things. With a minimal program just using curls url api the `https://xn--74h.com` becomes `https://☺.com/` but `https://www.xn--74h.com` stays the same. trurl behaves the same way for me (because its just wrapping the libcurl api). Are you seeing this behavior as well?

[...]

PR #393, opened 2025-05-15
trurl: use curl_free() on a pointer coming from curl_maprintf()

dfandrich:
One instance of a pointer coming from curl_maprintf() was freed with a
normal free() which breaks and causes ASAN/valgrind to complain when
using a debug-enabled libcurl. Since the function could return a normal
heap-allocated pointer or a curl-allocated one, copy the string and
curl_free() the original.

Closes #393

### DIFF
--- a/trurl.c
+++ b/trurl.c
@@ -1381,6 +1381,7 @@ static struct string *memdupzero(char *source, size_t len, bool *modified)
       int rlen;
       int leftside;
       int rightside;
+      char *temp;
 
       /* decode both sides */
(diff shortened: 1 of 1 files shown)


jacobmealey:
It might be worth adding the other idn domains from the ticket as tests.

dfandrich:
That would be worthwhile doing as part of the fix for #392, whatever that ends up being. That's not related to this issue, though.

jacobmealey:
ah my apologies, i looked at this right when i woke up. should have had some coffee first.

Issue #394, opened 2025-07-02
trurl's tests fail with curl 8.15-rc2


charles2910:
Hi,

This was already reported in #curl, but here seems like the best place. [Debian's CI caught a regression in trurl's tests](https://ci.debian.net/packages/t/trurl/unstable/amd64/61886411/#L55) with 8.15-rc2 that I uploaded yesterday. Here is the snippet:

<details><summary>Failed tests output</summary>
<p>

```
 24s 49: failed	--url https://curl.se/hello --set user=:hej:
 24s --- stdout --- 
 24s expected:
 24s 'https://%3ahej%3a@curl.se/hello\n'
 24s got:
 24s 'https://%3Ahej%3A@curl.se/hello\n'
 24s --- stderr --- 
 24s expected:
 24s ''
 24s got:
 24s ''
 24s --- returncode --- 
 24s expected:
 24s 0
 24s got:
 24s 0
 24s 128: failed	-s 'path=\\' -g '{path}\n{:path}' localhost
 24s --- stdout --- 
 24s expected:
 24s '/\\\\\n/%5c%5c\n'
 24s got:
 24s '/\\\\\n/%5C%5C\n'
 24s --- returncode --- 
 24s expected:
 24s 0
 24s got:
 24s 0
 24s --- stderr --- 
 24s expected:
 24s ''
 24s got:
 24s ''
 24s 129: failed	-s 'path=\\' --json localhost
 24s --- stdout --- 
 24s expected:
 24s [{'url': 'http://localhost/%5c%5c', 'parts': {'scheme': 'http', 'host': 'localhost', 'path': '/\\\\'}}]
 24s got:
 24s [{'url': 'http://localhost/%5C%5C', 'parts': {'scheme': 'http', 'host': 'localhost', 'path': '/\\\\'}}]
 24s --- returncode --- 
 24s expected:
 24s 0
 24s got:
 24s 0
 24s --- stderr --- 
 24s expected:
 24s ''
 24s got:
 24s ''
 24s 130: failed	-s 'path=\\' -g '{path}\n{:path}' --urlencode localhost
 24s --- stdout --- 
 24s expected:
 24s '/%5c%5c\n/%5c%5c\n'
 24s got:
 24s '/%5C%5C\n/%5C%5C\n'
 24s --- returncode --- 
 24s expected:
 24s 0
 24s got:
 24s 0
 24s --- stderr --- 
 24s expected:
 24s ''
 24s got:
 24s ''
[...]

PR #395, opened 2025-07-05
introduce 'excludes' option for tests

jacobmealey:
Addresses hexadecimal changes from curl 8.15. I've also duplicated the tests that were failing with the new uppercase hex values. 

Thanks @charles2910 for reporting this bug
fixes: #394

### DIFF
--- a/test.py
+++ b/test.py
@@ -250,11 +250,16 @@ def main(argc, argv):
                 print(f"Missing feature, skipping test {testIndex + 1}.")
                 numTestsSkipped += 1
                 continue
+            excludes = allTests[testIndex].get("excludes", None) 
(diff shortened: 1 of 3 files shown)


jacobmealey:
Not sure why the curl for win is failing. Will investigate tomorrow

jacobmealey:
The curl-for-win build is failing on my clone of trurl on master, which was synced this morning: https://github.com/jacobmealey/urler/actions/runs/16097961365/job/45423345923 
Both failing builds have the same error
```
ld.lld-19: error: undefined symbol: if_nametoindex
```

The daily builds on the curl-for-win repo are passing.

cc @vszakats

jacobmealey:
@vszakats that worked! Thank you!

PR #435, opened 2026-04-11
fixing strncmp length for imap protocol detection

arpitguptagithub:
The strncmp call comparing protocol names against "imap" used length 3 instead of 4. This would match any protocol starting with "ima", which is incorrect. The code should use 4 to correctly match "imap" (and its prefix in "imaps").


bagder (approved):
Yikes, how silly. Nice catch!

PR #436, opened 2026-04-11
Replace: add NULL checks for memdupzero/memdupdec return values

arpitguptagithub:
Both memdupzero() and memdupdec() can return NULL on allocation failure. The call site in replace() dereferenced them unconditionally, which would crash on out-of-memory. 

I think adding the NULL checks and call errorf() on failure, will make it consistent with the rest of the codebase.


bagder (approved):

bagder:
Thanks!

PR #437, opened 2026-04-11
Freeqpairs, fixing memry leak for trimmed/replaced query pairs

arpitguptagithub:
When trim() or replace() marks a query pair as deleted, they set len = 0 but allocate a new str via xstrdup(""). Since freeqpairs() only freed entries where len != 0, those strdup'd empty strings were leaked.

The fix is quite easy just check str (non-NULL) instead of len (non-zero) as the guard for freeing.

Add a test that trims a middle query key to exercise this code path.

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -3520,5 +3520,21 @@
       "returncode": 0,
       "stderr": ""
     }
+  },
+  {
+    "input": {
+      "arguments": [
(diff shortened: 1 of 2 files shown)


bagder (approved):

bagder:
Thanks!

PR #438, opened 2026-04-11
Fixing cmpfunc: fix sort to distinguish prefix-overlapping query keys

arpitguptagithub:
When one query key is a prefix of another (e.g. 'a' vs 'ab'), the comparison function returned 0 (equal), making --sort-query produce non-deterministic order for such pairs.

I fix it by returning the length difference when the common prefix matches, so shorter keys sort before longer ones.

Adding a test with prefix-overlapping query keys to verify deterministic sort order.

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -3536,5 +3536,18 @@
       "returncode": 0,
       "stderr": ""
     }
+  },
+  {
+    "input": {
+      "arguments": [
+        "https://example.com/?ab=2&a=1&abc=3",
+        "--sort-query"
+      ]
+    },
+    "expected": {
+      "stdout": "https://example.com/?a=1&ab=2&abc=3\n",
+      "returncode": 0,
+      "stderr": ""
+    }
   }
 ]
--- a/tests.json
+++ b/tests.json
@@ -764,6 +764,9 @@
         "query=user=me"
       ]
     },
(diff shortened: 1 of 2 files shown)


arpitguptagithub:
@bagder ig now this will be okay, have a look? please

bagder:
> This branch cannot be rebased due to conflicts

Please rebase your branch ontop of master and force-push

arpitguptagithub:
FYI

I have added  a regresiion test-`--sort-query` with prefix-overlapping keys (ab=2&a=1&abc=3) verifying correct sort output `(a=1&ab=2&abc=3).
`
and along with that also added a "excludes": ["uppercase-hex"] to 3 existing upstream tests (tests 52, 132, 172): These tests have lowercase hex in their expected output (%3d, %5c, %2f) but fail on CI builds linked against libcurl 8.20.0-DEV which produces uppercase hex (%3D, %5C, %2F) The output on the github Actions UI for test output made it very difficult to spot the actual difference. 

anyhow adding the excludes annotation skips them on uppercase-hex builds, matching the pattern already used by other tests in the repo (e.g. tests around line 717). 

These are pre-existing failures unrelated to the cmpfunc fix.

Hope thats fine.

PR #441, opened 2026-05-01
tests: mark some additional tests uppercase-hex sensitive

### LINKED ISSUE #440: More tests failing due to case sensitivity with curl 8.20.0
Hi,

We are seeing 3 more tests failing with curl 8.20.0 in Debian: https://ci.debian.net/packages/t/trurl/testing/amd64/70698738/#L244

```
 61s 52: failed	--url https://curl.se/hello --set query=user=me
 61s --- stdout --- 
 61s expected:
 61s 'https://curl.se/hello?user%3dme\n'
 61s got:
 61s 'https://curl.se/hello?user%3Dme\n'
[...]
 61s 132: failed	-s scheme:=http -s host:=localhost -s path:=/ABC%5C%5C -s 'query:=a&b&a%26b'
 61s --- stdout --- 
 61s expected:
 61s 'http://localhost/ABC%5c%5c?a&b&a%26b\n'
 61s got:
 61s 'http://localhost/ABC%5C%5C?a&b&a%26b\n'
[...]
 61s 72: failed	https://example.com/one/t%61o/%2F%42/ --append path=%61
 61s --- stdout --- 
 61s expected:
 61s 'https://example.com/one/tao/%2fB/%2561\n'
 61s got:
 61s 'https://example.com/one/tao/%2FB/%2561\n'
```

I've prepared a [patch](https://salsa.debian.org/debian/trurl/-/commit/52a2776583b8632e081c226e350cfdebf44211e4) (though it's a bit painful to have duplicated tests only with hex case difference :-( ) and will open a merge request.

And since we are here, could we have a 0.17.0 release? :p

charles2910:
Since 8.20.0, three more tests started to fail because of the case used to represent hex digits.

Fixes: #440

### DIFF
--- a/tests.json
+++ b/tests.json
@@ -764,8 +764,29 @@
         "query=user=me"
       ]
     },
+    "excludes": [
+      "uppercase-hex"
+    ],
     "expected": {
       "stdout": "https://curl.se/hello?user%3dme\n",
+      "stderr": "",
+      "returncode": 0
+    }
+  },
+  {
+    "input": {
+      "arguments": [
+        "--url",
+        "https://curl.se/hello",
+        "--set",
+        "query=user=me"
+      ]
+    },
+    "required": [
+      "uppercase-hex"
+    ],
+    "expected": {
+      "stdout": "https://curl.se/hello?user%3Dme\n",
       "stderr": "",
       "returncode": 0
     }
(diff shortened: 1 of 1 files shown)


jacobmealey:
I wonder if instead of trying to add a keyword every time libcurl changes the case of something, we instead add an `ignore_case` flag to the affected test cases? I had done to original work for the `uppercase-hex` work and at the time I thought it was a nice solution, but now I can see it getting a bit annoying.

vszakats:
I like the `ignore-case` flag idea, and/or perhaps make the comparison
recognize `%` encoded bytes and ignore case automatically?

jacobmealey:
That's even better, should we remove the `uppercase-hex` as a "feature" as well? 

Or should making percent encoded fields lower (or upper case) be part of trurls normalization step?

vszakats:
> That's even better, should we remove the `uppercase-hex` as a "feature" as well?
> 
> Or should making percent encoded fields lower (or upper case) be part of trurls normalization step?

I tend to think that passing it through as-is helps making sure that
consumer apps remain prepared for both casings, improving their
robustness along the way. It also avoid trurl forming an opinion on 
casing.

charles2910:
> perhaps make the comparison recognize % encoded bytes and ignore case automatically?

I agree that having a way to ignore the case of percent encoded stuff is the better solution here. Doing that would also allow us to drop the duplicate test cases for `uppercase-hex` and `lowercase-hex`. So I'll try to get back to this in the coming weeks and have another MR for it (things are a bit messy right now, so I expect to have free time to really work on it after the second week of june)

PR #442, opened 2026-05-16
GHA: explicitly `brew update` before `brew install`

vszakats:


vszakats:
Hm, rather skipping this due to slow performance when
there are updates. Chances to run into an issue as seen
in curl/curl is less likely here.

